PR #8, opened 2021-04-07
Update quickstart

flavio:
* Ensure the new CRD are being used
* Removed no longer valid examples, the pod-privileged policy doesn't
  have trusted and non-trusted users anymore
* Ensure MutatingWebhookConfiguration are covered by the quickstart too

### DIFF
--- a/src/quick-start.md
+++ b/src/quick-start.md
@@ -2,7 +2,7 @@
 
 The Kubewarden stack is made of the following components:
 
-  * An arbitrary number of `AdmissionPolicy` resources: this is how policies
+  * An arbitrary number of `ClusterAdmissionPolicy` resources: this is how policies
     are defined inside of Kubernetes
   * A Deployment of Kubewarden `policy-server`: this component loads all the
     policies defined by the administrators and evaluates them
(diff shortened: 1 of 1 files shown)


src/quick-start.md
```diff
@@ -113,15 +115,12 @@ spec:
...
 EOF
 ```
 
 This will produce the following output:
 ```console
-admissionpolicy.kubewarden.io/privileged-pods created
+clusteradmissionpolicy.kubewarden.io/privileged-pods created
 ```
 
 Defining the `AdmissionPolicy` will lead to a rollout of the Kubewarden Policy
```
ereslibre, line 126:
```suggestion
Defining the `ClusterAdmissionPolicy` will lead to a rollout of the Kubewarden Policy
```

ereslibre (changes requested):
Thank you @flavio, leftover I spotted

src/quick-start.md
```diff
@@ -67,6 +70,8 @@ This is a quick overview of the attributes of the `AdmissionPolicy` resource:
 * `operations`: what operations for the previously given types should
   be forwarded to this admission policy by the API server for
   evaluation.
+* `mutating`: a boolean value that must be set to `true` for policies that can
```
kkaempf, line 73: Shouldn't `mutating` flagged as `optional` ?

kkaempf (approved):

src/quick-start.md
flavio: We made it mandatory because, if a mutating policy is reported a validation only, the kubernetes API will choke on the mutation requests.

I'm not particularly happy about the amount of information a user has to write into our CR declaration, especially because many of these details can be embedded into the metadata of the `.wasm` policy file. We have plans to improved that once we have more time.

PR #24, opened 2021-04-14
Add architecture documentation

ereslibre:


src/architecture.md
```diff
@@ -0,0 +1,104 @@
...
+
+The `policy-server` will be the component responsible for refreshing
+this information when it considers, and the policy always retrieves
+tha latest available information exposed by the `policy-server`.
+
+## Satellite components
+
+Despite the `kubewarden-controller` and the `policy-server` are the
+main components, there are other satellite components worth
+mentioning. Let's go through them.
```
flavio, line 100: I would leave that out of this document. We can talk about that inside of the policy section of the docs where we already talk about the data types and functions used by host and guest policies to communicate

src/architecture.md
```diff
@@ -0,0 +1,104 @@
...
+
+### waPC: Wasm host (`policy-server`) and guest (`policy`) intercommunication
```
flavio, line 102: I would leave that out of this document. We can talk about that inside of the policy section of the docs where we already talk about the data types and functions used by host and guest policies to communicate

src/architecture.md
```diff
@@ -0,0 +1,104 @@
...
+
+### `policy-testdrive`
```
flavio, line 104: We already have a section about that on the docs

src/architecture.md
```diff
@@ -0,0 +1,104 @@
...
+this cached information.
+
+The result of these procedure calls is the JSON-encoded result of the
+list of resources, as provided by Kubernetes.
+
+The `policy-server` will be the component responsible for refreshing
+this information when it considers, and the policy always retrieves
+tha latest available information exposed by the `policy-server`.
+
+## Satellite components
```
flavio, line 96: I would drop this section entirely, see below

flavio (commented):
Looks like a good starting point. I can take it and dig deeper in some areas

src/architecture.md
```diff
@@ -0,0 +1,104 @@
...
+
+- [`kubewarden-controller`](https://github.com/kubewarden/kubewarden-controller):
+  this is the component that is deployed on top of the Kubernetes
+  cluster that you want to enrich with Wasm admission and mutation
+  policies.
+
+- [`policy-server`](https://github.com/kubewarden/policy-server):
+  its main responsibility is to instantiate the Wasm policies inside
+  safe Wasm sandboxes. A single `policy-server` can instantiate a
+  number of policies, each one running in a sandboxed environment,
```
kkaempf, line 19: "a number" sounds if it's limited. Either use "any number" or "many"

src/architecture.md
```diff
@@ -0,0 +1,104 @@
+# Architecture
+
+## Diagram
+
+TBD
+
+## Main components
+
```
kkaempf, line 8: May start with the fact that kubewarden is a custom resource and kubewarden-controller is the implementation of this resource. It starts 'policy-server' as a helper to provide the safe Wasm sandbox.

src/architecture.md
```diff
@@ -0,0 +1,104 @@
...
+  its main responsibility is to instantiate the Wasm policies inside
+  safe Wasm sandboxes. A single `policy-server` can instantiate a
+  number of policies, each one running in a sandboxed environment,
+  lacking access to other policies execution environment.
+
+### `kubewarden-controller`
+
+The `kubewarden-controller` is deployed on top of the target
+Kubernetes cluster and its main duty is to reconcile
+`ClusterAdmissionPolicy` resources.
```
kkaempf, line 26: Hmm, quick-start still talks about "AdmissionPolicy" at multiple places.

src/architecture.md
kkaempf: Also a brief introduction to (Kubernetes) policies in general (policies are registered with k8s, multiple policies can be chained, policy passes only if all in the chain pass, etc.) Then link to https://kubernetes.io/docs/concepts/policy/

kkaempf:
I'd like to see the architecture description enriched with pointer into the code, esp. naming respective entry points (functions) for the various steps. Kind of guideline for code contributors.

Is there upstream documentation of how to write policy implementations in k8s ?

PR #28, opened 2021-04-21
Added draft version for the context aware policies

flavio:
This is some temporary data, we need to dig deeper into the topic


ereslibre (approved):
Looks perfect to me as a first approach, thank you for taking care of this.

PR #29, opened 2021-04-21
Edit page feature

flavio:
Testing the new updated deployment action. This adds a "edit page" link on top of each page


flavio:
Everything seems to be working as expected, merging

PR #37, opened 2021-06-07
Explain that cargo generate must be installed

charlieegan3:
cargo generate is not part of a default rust installation so I have attempted to explain how to install the required tool


ereslibre (approved):
LGTM, thank you @charlieegan3

PR #40, opened 2021-06-11
Adapt documentation

ereslibre:
- Replace `policy-testdrive` with `kwctl` and adapt their command
names, arguments, workflow...
- Replace `wasm-to-oci` with `kwctl`
- Update the go project template so it includes some expected labels
to execute when following this documentation
- Make sure that all examples work as expected

Fixes: https://github.com/kubewarden/kubewarden.io/issues/15


flavio (approved):
Thanks for all the fixes!

PR #41, opened 2021-06-20
Add the namespace to the uninstall command

y-taka-23:
The release is installed in the `kubewarden` namespace, thus you have to specify it for uninstallation too.

> helm install --namespace kubewarden --create-namespace kubewarden-controller kubewarden/kubewarden-controller


flavio:
Thanks for the fix!

Issue #53, opened 2021-09-30
Update installation documentation for the new CRD helm chart


raulcabello:
- update quick start install instructions
- update quick start uninstall instructions

flavio:
fixed

PR #54, opened 2021-09-30
Update documentation for the new CRD helm chart

raulcabello:
fixes kubewarden/docs#53

### DIFF
--- a/src/quick-start.md
+++ b/src/quick-start.md
@@ -18,12 +18,16 @@ Currently, the chart depends on cert-manager. Make sure you have [`cert-manager`
 ```console
 helm repo add kubewarden https://charts.kubewarden.io
 kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.5.3/cert-manager.yaml
-helm install --wait --namespace kubewarden --create-namespace kubewarden-controller kubewarden/kubewarden-controller
+helm install --create-namespace -n kubewarden kubewarden-crds kubewarden/kubewarden-crds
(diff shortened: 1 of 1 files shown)


src/quick-start.md
```diff
@@ -18,12 +18,16 @@ Currently, the chart depends on cert-manager. Make sure you have [`cert-manager`
 ```console
 helm repo add kubewarden https://charts.kubewarden.io
 kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.5.3/cert-manager.yaml
-helm install --wait --namespace kubewarden --create-namespace kubewarden-controller kubewarden/kubewarden-controller
+helm install --create-namespace -n kubewarden kubewarden-crds kubewarden/kubewarden-crds
```
flavio, line 21: I don't think you have to create a namespace here, this is just installing the CRDs. Obviously it won't hurt, but maybe we should move it to the next line?

flavio (approved):
LGTM

src/quick-start.md
raulcabello: it will be installed in the default namespace if we don't provide one. So I think we should use the kubewarden ns

ereslibre: CRD's are cluster wide though, so it should not matter

ereslibre (approved):

jvanz (approved):

src/quick-start.md
flavio: Good point

PR #56, opened 2021-10-01
Add section explaining how to enable tracing

flavio:
Add a new "operator manual" section. Under this section we can add documentation explaining how to perform advanced tuning of Kubewarden.

As a starting point, a document explaining how to enable tracing support is created.

> *NOTE WELL:* this branch builds on top of the `new-architecture` branch


viccuad (approved):
LGTM, but I haven't tested the steps yet.

src/operator-manual/tracing/01-quickstart.md
```diff
@@ -0,0 +1,418 @@
+# Tracing
+
+This section illustrates how to enable tracing support of
+Policy Server.
+
+Tracing allows to collect fine grained details about policy evaluations. It can
+be a useful tool for debugging issues inside of your Kubewarden deployment.
```
ereslibre, line 7:
```suggestion
be a useful tool for debugging issues inside of your Kubewarden deployment and policies.
```

src/operator-manual/tracing/01-quickstart.md
```diff
@@ -0,0 +1,418 @@
...
+The OpenTelemetry Operator requires [cert-manager](https://cert-manager.io/docs/installation/)
+to be installed inside of the cluster.
+
+This can be done with this command:
+
+```console
+kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.5.3/cert-manager.yaml
+```
+
+Once cert-manager is up and running, the operator can be installed in this way:
```
ereslibre, line 164: I think we can add also a command to make sure it's running `kubectl wait --for=condition=Available deployment --timeout=2m -n cert-manager --all`.

src/operator-manual/tracing/01-quickstart.md
```diff
@@ -0,0 +1,418 @@
...
+      operator: In
+      values: ["default"]
+  mutating: false
+EOF
+```
+
+This policy will enforce all the Kubernetes resources declared inside of the
+`default` Namespace to have a label called `owner`.
+
+Once the policy is active, we can try it out in this way:
```
ereslibre, line 322: We could wait with `kubectl wait --for=condition=PolicyActive clusteradmissionpolicy/safe-labels`

flavio: Added, thanks!

src/operator-manual/tracing/01-quickstart.md
flavio: Added, thanks

ereslibre (approved):

raulcabello (approved):
LGTM I just followed the documentation and I can see Jaeger traces in my local machine.
The only thing I found is that we are using the same jaeger service name for all policy servers https://github.com/kubewarden/policy-server/blob/main/src/cli.rs#L13 It is not updated for the new architecture. We should probably use the `PolicyServer` name, so we can see traces if we have more than one servers.
But this is not related to the documentation, so I'll create a new card for this.

src/operator-manual/tracing/01-quickstart.md
```diff
@@ -0,0 +1,430 @@
...
+        - CREATE
+        - UPDATE
+  namespaceSelector:
+    matchExpressions:
+    - key: kubernetes.io/metadata.name
+      operator: In
+      values: ["default"]
+  mutating: false
+EOF
+```
```
raulcabello, line 323: I cannot see any resource in https://github.com/derailed/k9s after applying this policy. It works fine with kubectl, so probably k9s is initializing resources without the owner label.  I don't think there is a problem with the policy itself, but for k9s users cluster seems to be broken as they cannot list any resource at all.
Error:
`│                           < Ruroh? admission webhook "safe-labels.kubewarden.admission" denied the request: The following mandatory labels are missing: owner >                            │
`

src/operator-manual/tracing/01-quickstart.md
```diff
@@ -0,0 +1,430 @@
...
+Then we can proceed with the installation of the helm chart:
+
+```console
+helm install --wait --namespace kubewarden --values values.yaml kubewarden-controller kubewarden/kubewarden-controller
+```
+
+This leads to the creation of the `default` instance of `PolicyServer`:
+
+```console
+kubectl get -n kubewarden policyservers.policies.kubewarden.io -n kubewarden
```
jvanz, line 281: Duplicates `-n` flag?

flavio:
> The only thing I found is that we are using the same jaeger service name for all policy servers https://github.com/kubewarden/policy-server/blob/main/src/cli.rs#L13 It is not updated for the new architecture. We should probably use the PolicyServer name, so we can see traces if we have more than one servers.
But this is not related to the documentation, so I'll create a new card for this.

@raulcabello I thought about that as well. I don't think we should change the service name. There's some code already merged that adds a `hostname` tag to each trace event as a tag. By doing that you can filter traces. 

The hostname, on kubernetes, is calculated starting from the name of the Deployment.

src/operator-manual/tracing/01-quickstart.md
flavio: Good catch. I've fixed that

raulcabello:
ok @flavio , filtering by the `hostname` tag sounds good. Thanks!

src/operator-manual/tracing/01-quickstart.md
flavio: I think each Kubernetes distro is doing its thing with automatic labels. I was running the original example against minikube, but it turns out k3s was not adding this label.

I've changed the example to use different key/values and create ad-hoc Namespace objects

src/operator-manual/tracing/01-quickstart.md
```diff
@@ -0,0 +1,458 @@
...
+    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
+🌟  Enabled addons: storage-provisioner, default-storageclass
+🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default
+```
+
+Now we have to enable the Ingress addon:
+
+```console
+minikube addons enable ingress
+```
```
jvanz, line 64: This is nothing that prevents merging the request but just a comment just in case you do not know. We can install the addons in the minikube start command with the `--addons` flag. ;)

jvanz (approved):
LGTM!

src/operator-manual/tracing/01-quickstart.md
flavio: Nice, I didn't know about that

flavio:
I'm going to publish that in a couple of days, I don't want to step over the blog post about the new architecture

PR #60, opened 2021-10-13
Add Google analytics to docs website

kkaempf:
Signed-off-by: Klaus Kämpf <kkaempf@suse.de>


ereslibre:
> It's the same (javascript) code as used in www.kubewarden.io afaik

The mdBook one is missing one attribute that tells Google to anonymize IP's: `ga('set','anonymizeIp',!0)`. This can be found on the JS generated for kubewarden.io, but I don't think is present in the generated JS code for mdBook:

https://github.com/rust-lang/mdBook/blob/a72d6002b78388885fbb879724bdfadc2ffb776c/src/theme/index.hbs#L246-L254

flavio:
@kkaempf let me know if you need help with the PR against mdbook. I've already done some fixes there

Issue #61, opened 2021-10-15
The `Quick Start` documentation has references to the recently removed Jaeger tracing. 


jvanz:
The `Quick Start` documentation has references to the recently [removed Jaeger tracing](https://github.com/kubewarden/policy-server/pull/104). We need to remove the references or update to Open Telemetry configuration.

https://docs.kubewarden.io/quick-start.html#policy-server

PR #66, opened 2021-11-03
quickstart: remove jaeger mention

ereslibre:
Fixes: https://github.com/kubewarden/docs/issues/61

For the quick start there is no need to add any integration with
Jaeger or OpenTelemetry.


viccuad (approved):
:rocket: !

PR #73, opened 2021-11-26
Updated quick-start.md

nunix:
Global update of the `quick-start` page.

Inserted tables for a better visibility of the different options and reviewed the examples.

Certain terminology might be better suited, and will be incorporated upon reviews feedback.

PR #74, opened 2021-11-26
fix quickstart page

chrisns:
ghcr.io/kubewarden/policy-server:v1.0.0 doesn't exist use latest for now

### DIFF
--- a/src/quick-start.md
+++ b/src/quick-start.md
@@ -153,7 +153,8 @@ kind: ClusterAdmissionPolicy
 metadata:
   name: privileged-pods
 spec:
-  module: registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.5
+  policyServer: reserved-instance-for-tenant-a
+  module: registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9
   rules:
   - apiGroups: [""]
     apiVersions: ["v1"]
--- a/src/quick-start.md
+++ b/src/quick-start.md
@@ -50,7 +50,6 @@ metadata:
 spec:
   image: ghcr.io/kubewarden/policy-server:latest
   replicas: 2
-  serviceAccountName: sa
   env:
   - name: KUBEWARDEN_LOG_LEVEL
     value: debug
--- a/src/quick-start.md
+++ b/src/quick-start.md
@@ -48,7 +48,7 @@ kind: PolicyServer
 metadata:
   name: reserved-instance-for-tenant-a
 spec:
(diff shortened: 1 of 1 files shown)


ereslibre (approved):
Thank you, although is just an example of a PolicyServer resource, no need to refer to a non-existant tag.

chrisns:
> Thank you, although is just an example of a PolicyServer resource, no need to refer to a non-existant tag.

👍 It's the quick start guide, not an 'example'? so you expect to be able to copy+paste to success :)

Tangentially: I'm currently working on https://github.com/appvia/psp-migration which would be great if you've got any time/interest to collaborate on?

ereslibre:
> Tangentially: I'm currently working on https://github.com/appvia/psp-migration which would be great if you've got any time/interest to collaborate on?

Thank you for the heads up. I'm going to bring this up with the team for sure.

src/quick-start.md
```diff
@@ -153,7 +153,8 @@ kind: ClusterAdmissionPolicy
 metadata:
   name: privileged-pods
 spec:
-  module: registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.5
+  policyServer: reserved-instance-for-tenant-a
```
ereslibre, line 156: I would not change this here. The original structure of the quickstart was:

- Install
- Describe what a PolicyServer resource is
- Describe what a ClusterAdmissionPolicy resource is
- Enforce your first policy (without targeting any specific PolicyServer)

Given this page is going an overhaul in https://github.com/kubewarden/docs/pull/73/files, I think it's fine to fix the broken example, but changing this here requires deploying the PolicyServer described before, whereas we wanted to show this deployment as simple as possible (no need to specify a policyServer, it will be defaulted if missing).

I'll defer to @nunix for further approval/merge, given he's working on this overhaul.

nunix: It's a bit a caveat here as, in terms of example, running the commands will work (I did it several times the past days).

However I kept the `yaml` examples as is, as they were used "only" to explain the options inside (transformed in a table in the overhaul).
So while the `yaml` examples are there, in the quick start they are not used.

Still, I totally get the versioning problem and in this case, would you prefer to transform it into something generic (again, it's only used for explanations), such as: `module: registry://ghcr.io/kubewarden/policies/pod-privileged:<version tag>`

Hope this makes sense.

chrisns:
marking this PR as `draft`, will fix when we've figured out #75

flavio:
Closing the quickstart docs got changed a lot since the issue was opened. Now we have a version being referenced by the snippet

PR #81, opened 2022-02-01
sources.yaml default path for kwctl

raulcabello:
Fixes https://github.com/kubewarden/kwctl/issues/144
I just tested on Linux. I took the [official documentation](https://docs.rs/directories/3.0.2/directories/struct.ProjectDirs.html#method.config_dir) as a base to calculate the path for Windows and Mac. For the [parameters](https://docs.rs/directories/3.0.2/directories/struct.ProjectDirs.html#parameters) we are using `io.kubewarden` as `qualifier` and `kubewarden` as `application`


ereslibre:
Self-requesting a review because I am going to confirm the Mac path works as documented on the crate docs.

ereslibre (approved):
Checked on Linux, Mac and Windows. It works, thanks @raulcabello!

jvanz (approved):

PR #82, opened 2022-02-03
New mutating policies page

nunix:
Created the mutating policies validation page, based on the new Kubewarden controller and policy server versions.

Updated the Summary page to add it as a follow-up to the tasks page

### DIFF
--- a/src/mutating-policies.md
+++ b/src/mutating-policies.md
@@ -1,4 +1,4 @@
-### Mutating policies validation
+### Mutating policies
 
 Since [Kubewarden controller](https://github.com/kubewarden/kubewarden-controller) version v0.4.5 and [Policy Server](https://github.com/kubewarden/policy-server) version v0.2.6, the `mutating` field impacts how mutated requests are validated. If you deploy a policy which mutates requests and the `mutating` field set to `true`, the requests will act as a validated ones.
 
--- a/src/mutating-policies.md
+++ b/src/mutating-policies.md
@@ -1,8 +1,8 @@
 ### Mutating policies
 
-Since [Kubewarden controller](https://github.com/kubewarden/kubewarden-controller) version v0.4.5 and [Policy Server](https://github.com/kubewarden/policy-server) version v0.2.6, the `mutating` field impacts how mutated requests are validated. If you deploy a policy which mutates requests and the `mutating` field set to `true`, the requests will act as a validated ones.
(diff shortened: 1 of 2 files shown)


src/mutating-policies.md
```diff
@@ -0,0 +1,129 @@
+### Mutating policies validation
```
viccuad, line 1: Nitpicking:

```suggestion
### Mutating policies
```
(or something similar) to remove ambiguity; so people don't confuse that we are validating the yaml of the policy definition.

src/mutating-policies.md
```diff
@@ -0,0 +1,129 @@
...
+
+Since [Kubewarden controller](https://github.com/kubewarden/kubewarden-controller) version v0.4.5 and [Policy Server](https://github.com/kubewarden/policy-server) version v0.2.6, the `mutating` field impacts how mutated requests are validated. If you deploy a policy which mutates requests and the `mutating` field set to `true`, the requests will act as a validated ones.
```
viccuad, line 3: I would rephrase the paragraph to be clearer and shorted. Something of the like:
```suggestion
A mutating policy is one that can mutate requests sent to it. This means that it will change the objects when instantiating them. Those policies are written so they rebuild the request to be conformant with the policy. If you want them to instantiate the rebuilt request, that is, to make a policy behave as mutating, set `ClusterAdmissionPolicy.mutating` to `true`.

However, if you you just want the policy to validate the requests without mutating them, as a normal non-mutating policy, you can always set `ClusterAdmissionPolicy.mutating` to `false`.

Here's an example. When setting to true: (queue to psp-user-group example)
```

I would drop the "since kubewwarden-controller ..", we can assume that the docs target latest versions, if not we are going to have a lot of "since". In the future we could move to versioned docs instead.

Then later I would explain the behaviour with `ClusterAdmissionPolicy.mutating` set to `false`, and drop the "However, if you set the `mutating` field to false.." until after the example that sets it to false.

jvanz (approved):
I agree with the Victor's comments. Besides that, LGTM

src/mutating-policies.md
nunix: Updated the first sentence ([bea3340](https://github.com/kubewarden/docs/pull/82/commits/bea3340a01e5f4998a5276438b2ab27064be6dbc))

viccuad (approved):
LGTM! Many thanks!

PR #94, opened 2022-03-09
quickstart: Use `PolicyServer` instead of `Policy Server`

viccuad:
## Description

Use `PolicyServer` instead of `Policy Server`, to match the Kubernetes resource name.

### DIFF
--- a/src/quick-start.md
+++ b/src/quick-start.md
@@ -47,11 +47,11 @@ Kubewarden has three main components which you will interact with:
 * The ClusterAdmissionPolicy
 * The AdmissionPolicy
 
-### Policy Server
+### PolicyServer
 
-A Kubewarden Policy Server is completely managed by the `kubewarden-controller` and multiple Policy Servers can be deployed in the same Kubernetes cluster.
+A Kubewarden `PolicyServer` is completely managed by the `kubewarden-controller` and multiple `PolicyServers` can be deployed in the same Kubernetes cluster.
 
-The Policy Server is the component which executes the Kubewarden policies when requests arrive and validates them.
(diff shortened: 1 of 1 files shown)


src/quick-start.md
```diff
@@ -47,11 +47,11 @@ Kubewarden has three main components which you will interact with:
 * The ClusterAdmissionPolicy
 * The AdmissionPolicy
 
-### Policy Server
+### PolicyServer
 
-A Kubewarden Policy Server is completely managed by the `kubewarden-controller` and multiple Policy Servers can be deployed in the same Kubernetes cluster.
+A Kubewarden `PolicyServer` is completely managed by the `kubewarden-controller` and multiple `PolicyServers` can be deployed in the same Kubernetes cluster.
```
divya-mohan0209, line 52: I’m not sure we should be capitalising the pluralised version.

Possibly, we can rephrase it to:

```suggestion
A Kubewarden `PolicyServer` is completely managed by the `kubewarden-controller` and multiple instances of the `PolicyServer` can be deployed in the same Kubernetes cluster.
```

divya-mohan0209 (commented):
Grammatical nit. Not a blocker, but it would read better.

src/quick-start.md
viccuad: Using the plural is ok, the Kubernetes CRD defines singular and plural names:
https://github.com/kubewarden/kubewarden-controller/blob/e0433fc3774d06dcf5e08bf2c600ad0117b89448/config/crd/bases/policies.kubewarden.io_policyservers.yaml#L15

divya-mohan0209: Cool, then LGTM! Sorry :)

viccuad: thanks! merging

PR #104, opened 2022-04-21
Dev docusaurus

bisht-richa:
## Description

Fix #XXX

## Test


viccuad (commented):
Gave it a spin locally, it's wonderful, looking forward for it to land. I suppose it will land after the other PRs.
Maybe it makes sense to separate commits more granularly, into a commit that adds new files, a commit that changes filepath of existing content, and a commit that changes things inside of the new content. That way, a `git rebase main` from this branch may be easier. But this is all personal preference for updating this PR down the line.

jvanz (commented):
I just ran the code locally and the new version is great! 

I have only one question. Should we push the `node_modules` directory ?  It seems something that should not be pushed to the remote.  I think the developers can call npm install at the first time and move on. No need to commit this. I'm asking this because in my local env I had to remove the directory and call `npm install` again. 

However, if  this is a best practice in the npm world or if you will do more changes, ignore this comment. ;)

docusaurus.config.js
```diff
@@ -0,0 +1,112 @@
...
+module.exports = {
+  title:
+    "The open source hyperconverged infrastructure (HCI) solution for a cloud native world",
+  tagline: "",
+  url: "https://kubewarden.io",
+  baseUrl: "/",
+  onBrokenLinks: "throw",
+  onBrokenMarkdownLinks: "warn",
+  favicon: "img/favicon.ico",
+  organizationName: "harvester", // Usually your GitHub org/user name.
```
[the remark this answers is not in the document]
divya-mohan0209: Addressed in the latest set of commits :)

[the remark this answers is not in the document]
divya-mohan0209: Also - no, node_modules aren't required. I just pushed it by mistake. Removed it in the latest set :)

docs/tasks.md
```diff
@@ -17,7 +17,7 @@ The Kubewarden Policy Hub hosts policies contributed by the community. For examp
 
 As shown in the picture below, once you find the policy to be tested, you can copy the registry path<sup>1</sup> or download<sup>2</sup> the `Wasm` binary containing the policy and additional metadata:
 
-![Kubewarden Policy Hub](/img/tasks-policy-hub.png)
+![Kubewarden Policy Hub](/images/tasks-policy-hub.png)
```
raulcabello, line 20: I can't build it because it doesn't find this image

raulcabello (commented):
I run it locally and I love it!! I had a few problems building the site:

```
SyntaxError: /home/raul/Development/kubewarden/docs-1/docs/writing-policies/spec/04-mutating-policies.md: Unexpected token, expected "}" (139:11)
````

To fix it I had to remove this [table](https://github.com/kubewarden/docs/pull/104/files#diff-b6047823b91cbfb5d2ff5f7e708ec6c78c64122dda8730ac318498af7ebeeef2L140-L202)

Also I had to remove [this image](https://github.com/divya-mohan0209/docs-1/blob/dev-docusaurus/docs/tasks.md?plain=1#L20), as it couldn't be found. 

Not sure if I'm doing something wrong or I'm missing something here, I just run `yarn install` and then `yarn start`

**Comment by ereslibre** [comment]
Thank you! It looks beautiful and well organized. I have the same couple of issues @raulcabello points to. I'm also running it through `yarn install` and `yarn start`.

I went through all the pages and I found some small issues, not sure how you want them to be reported.

**Comment by divya-mohan0209** [comment]
The issue with the mutating bit is because I re-added it

File: README.md [hunk 3]
```diff
@@ -1,2 +1,33 @@
-This repo contains the source of the Kubewarden documentation, the rendered manual
-can be found [here](https://docs.kubewarden.io/).
+# harvesterhci.io
```
jvanz, line 1: Update Harvester reference. ;)

docusaurus.config.js
```diff
@@ -0,0 +1,112 @@
...
+      items: [
+        {
+          type: "doc",
+          docId: "introduction",
+          position: "right",
+          label: "Docs",
+          className: "navbar__docs",
+        },
+        {
+          href: "https://github.com/harvester/harvester",
```
jvanz, line 37: Update Harvester reference. ;)

package.json
```diff
@@ -0,0 +1,43 @@
...
+      "not op_mini all"
+    ],
+    "development": [
+      "last 1 chrome version",
+      "last 1 firefox version",
+      "last 1 safari version"
+    ]
+  },
+  "main": "index.js",
+  "repository": "https://github.com/harvester/harvester.github.io.git",
```
jvanz, line 40: Update Harvester reference. ;)

jvanz (changes requested):

flavio (approved):

ereslibre (commented):
It looks amazing, thank you! Some comments:

- Here there's a command that has to be boxed (at the end of the screnshot)

Screenshot 2022-05-06 at 09 08 48

- I think we could pretty format and color YAML files where applicable

Screenshot 2022-05-06 at 09 09 45

I understand that we cannot do so when YAML files are embedded in a `kubectl apply -f -` command, but that's fine, I only mean in this case, where the YAML file appears alone.

- Along the lines of the previous point, does docusaurus support Rust highlighting?

Screenshot 2022-05-06 at 09 11 11

- I think the protocol contracts would be more readable with a monospaced font

Screenshot 2022-05-06 at 09 10 32

Screenshot 2022-05-06 at 09 10 49

Thank you, I love this version!

viccuad:
> I understand that we cannot do so when YAML files are embedded in a kubectl apply -f - command, but that's fine, I only mean in this case, where the YAML file appears alone.

That falls on me, I convinced José to change all the markdown blocks in that page to ` ```console`, and indeed some of them could be ` ```yaml`, including the ones that start with `kubectl apply -f`. It would look better.
See: https://github.com/kubewarden/docs/blob/main/src/psp-migration.md?plain=1#L50

viccuad (approved):

raulcabello (approved):
LGTM, thanks @divya-mohan0209 !

PR #114, opened 2022-05-13
Warns user about policy server version and the `sa` field.

jvanz:
Updates the Quick Start guide changing the default PolicyServer version and warning the user about the `serviceAccountName` which can cause the deployment to fail.

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -85,13 +85,15 @@ kind: PolicyServer
 metadata:
   name: reserved-instance-for-tenant-a
 spec:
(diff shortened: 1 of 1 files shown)


raulcabello (approved):
LGTM, thanks @jvanz !

viccuad:
how about removing the ServiceAccount for the quickstart? maybe it's cleaner, and it's just the quickstart. Or, set it as `serviceAccount: ~` if it works, so people know that they can use it, yet, its unset for this example.

viccuad (approved):
LGTM, thanks!

raulcabello (approved):
LGTM

Issue #111, opened 2022-05-06
Docusaurus nits


divya-mohan0209:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Issue opened to track nits mentioned by @ereslibre in [this comment](https://github.com/kubewarden/docs/pull/104#pullrequestreview-964265371) on the PR #104 

### Environment

```markdown
- OS:
- Architecture:
```

flavio:
Temporarily blocked because @divya-mohan0209 is busy with KubeCon and travels

flavio:
@divya-mohan0209 I've addressed the issues while working on the changes I've done to the Go SDK. I'll open a PR ASAP. In the meantime, don't waste your time on that

flavio:
We can close this one, all the changes have been merged and deployed now

PR #122, opened 2022-06-07
Fix nits and extend wapc specs

flavio:
This PRs performs the following changes:

* feat: add documentation about the new host capabilities functions that have been introduced
* fix: sidebar: move all the topics about writing policies under the same top level entry
* feat: enable theme switching - allow users to toggle between dark and light mode
* fix: fix syntax highlighting of some yaml sections, plus add support for Rust and Rego. This solves https://github.com/kubewarden/docs/issues/111
* fix: fix code blocks inside of the waPC spec section. This solves https://github.com/kubewarden/docs/issues/111


docs/writing-policies/spec/host-capabilities/01-intro-host-capabilities.md
```diff
@@ -0,0 +1,39 @@
+---
+sidebar_label: "Host Capabilities Specification"
+title: ""
+---
+
+# Host capabilities specification
+
+While being evaluated, Kubewarden policies can access extra capabilities offered
+by the host environment.
+This mechanism works using an approach similar to traditional [RPC](https://en.wikipedia.org/wiki/Remote_procedure_call).
```
divya-mohan0209, line 10:
```suggestion
This mechanism uses an approach similar to traditional [RPC](https://en.wikipedia.org/wiki/Remote_procedure_call).
```

docs/writing-policies/spec/host-capabilities/03-container-registry.md
```diff
@@ -0,0 +1,49 @@
+---
+sidebar_label: "Container Registry Capabilities"
+title: ""
+---
+
+# Container registry capabilities
+
+Container registries can be used to distribute many types of OCI objects, going
+from well-known container images to generic OCI Artifacts. OCI Artifacts are used
+to store objects as Kubewarden Policies, Helm charts and more.
```
divya-mohan0209, line 10:
```suggestion
to store objects such as Kubewarden Policies, Helm charts, and more.
```

docusaurus.config.js
```diff
@@ -12,13 +12,20 @@ module.exports = {
...
+      ],
+    },
     colorMode: {
       // "light" | "dark"
       defaultMode: "light",
 
       // Hides the switch in the navbar
       // Useful if you want to support a single color mode
-      disableSwitch: true,
+      disableSwitch: false,
```
divya-mohan0209, line 28: I'd rather we keep this as true since we don't have the theming for dark mode yet.
```suggestion
      disableSwitch: false,
```

divya-mohan0209 (commented):
Overall LGTM, however can you paste a screenshot of how it looks when you build it locally?

I'm trying to see how we can integrate Netlify so that we can have a deploy preview.

docusaurus.config.js
flavio: Everything is working fine an looking natural here. I would enable this option, we can always tune the theme later on if needed (but right now I personally don't see the need)

# Dark mode

image

# Light mode

image

docs/writing-policies/spec/host-capabilities/03-container-registry.md
```diff
@@ -0,0 +1,49 @@
...
+
+These are the capabilities exposed by the Kubewarden host to interact with
```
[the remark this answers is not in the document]
flavio: No, that's not the intention. The first paragraph is just a description about what OCI registries are, I tried to highlight that they are not just about storing container images, other things can be added.

The next paragraph is more about: look, the remaining parts of this page are going to document the kind of capabilities that we make available to interact with container registries. Right now we have only one function available, maybe more will be added in the future.

These are the generic ideas behind these 2 paragraphs. I'm open to changes if these makes things easier to understand :)

flavio:
> Overall LGTM, however can you paste a screenshot of how it looks when you build it locally?
> 
> I'm trying to see how we can integrate Netlify so that we can have a deploy preview.

I've updated the PR description to include screenshots of the parts that have been changed

docs/writing-policies/spec/host-capabilities/03-container-registry.md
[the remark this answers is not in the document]
divya-mohan0209: So, in that case I think we'll need to rephrase the beginning of the second paragraph slightly differently.

```suggestion
Below documented are the capabilities exposed by the Kubewarden host to interact with
```

docusaurus.config.js
divya-mohan0209: I don't have any preference since I was told by the UI/UX team that the theming was only for light mode. If y'all are happy, I am good too!

divya-mohan0209 (changes requested):
Just one minor rephrasing, rest LGTM

docs/writing-policies/spec/host-capabilities/03-container-registry.md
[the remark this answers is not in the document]
flavio: That's perfect, thanks for the suggestion!

jordojordo (changes requested):
One small nitpick - I think there should be consistent spacing between the Github button and theme selector. If we add `20px` to the right side as well this should fix the issue:

thing1

I've added some styles locally which adds the margin depending on the size screen and this seems to work:
```css
/* ./src/css/custom.css */

@media (min-width: 996px) {
  .icon-github.navbar__github {
    margin-right: 20px;
  }
}
```

https://user-images.githubusercontent.com/40806497/172855324-ca155e08-1bcb-4151-a1db-a09c28181e17.mp4

viccuad (approved):
LGTM, thanks Flavio!

PR #137, opened 2022-08-11
Update the Rust instructions

flavio:
We've recently changed our rust compilation target to be `wasm32-wasi` instead of the `wasm32-unknown-unknown`.
This is required to prevent a Rust panic to happen when the policy is loaded.

For more details, see https://github.com/kubewarden/rust-policy-template/pull/25


viccuad (approved):

divya-mohan0209 (approved):
Lgtm!

viccuad:
merging then!

PR #138, opened 2022-08-24
Fix typo annotation -> annotations

kravciak:
```
~ kubectl logs -n kubewarden policy-server-default-bf9dc996b-9z96j
Cannot create sigstore verification config: Not a valid configuration file: Err(Message("unknown field `annotation`, expected one of `owner`, `key`, `annotations`", None))
```


viccuad (approved):
thanks!

PR #143, opened 2022-10-25
fix: update quickstart references

flavio:
* Use stable versions of our CRDs
* Use latest versions of the policies being demoed
* Update some of the output that changed over the time


divya-mohan0209 (approved):
LGTM

PR #160, opened 2023-01-05
Add reminder that kwctl needs to be authenticated

viccuad:
## Description

`kwctl` needs to be authenticated against the registry, just as `docker`.

Wrap command lines so it's more readable.

## Test


divya-mohan0209 (approved):
Lgtm, thanks!

jvanz (approved):

PR #163, opened 2023-01-11
Sidebar rework based on Divio paradigm

nunix:
As for the other SUSE Rancher projects, the Divio paradigm will be implemented for Kubewarden docs.

This change will be done in different steps and should bring an easier navigation for the docs.

This first step is a complete rework on the sidebar, to align it with the Divio paradigm.

The docs will be re-arranged under the four main groups, however if needed, additional groups will be created.

Signed-off-by: Nuno do Carmo <nuno.carmo@suse.com>


flavio:
I very much like it :clap:

nunix:
> Me too! In addition, I would like to move everything under `Root -> How Tos -> Operator Manual -> Reference Documentation` directly under `Root -> Reference Documentation` (example, [verification config](https://deploy-preview-163--silly-bunny-8cedd0.netlify.app/operator-manual/verification-config)).
> 
> I also would like to move the definitions of `sources.yaml` file in `Root -> How Tos -> Custom Certificate Authorities` under `Root -> Reference Documentation`.

The Operator Manual references has been moved to References as suggested.

For the `sources.yaml` file, as the content is very short we can either create a new page and put it under References as suggested, or if there's a complete example we can import it "as is" in a new page.

viccuad:
Thanks!
I would go further and move what is now under `Root -> Reference Documentation -> Operator Manual -> Foo` into `Root -> Reference Documentation -> Foo` if that makes sense. At least I would like to move the CRDs under  `Root -> Reference Documentation -> CRDs`, because that seems like one of the most accessed for newcomers.

For `sources.yaml`, true, is a bit weird being the content so short and a mix of reference and example. I suppose you mean something similar to the following:
- Rewrite  `HowTos ->  Custom Certificate Authorities` to be more how-to focused, saying to use a `sources.yaml` file for kwctl and PolicyServer, what behaviour does that bring, and provide just an example.
- This `HowTos ->  Custom Certificate Authorities`  points to the `Root -> Reference Documentation -> sources.yaml` for more information on settings besides the provided example.

jvanz:
I really like the new structure. :)

sidebars.js
```diff
@@ -1,52 +1,18 @@
 module.exports = {
   docs: [
     'introduction',
-    'quick-start',
+    `security/disclosure`,
```
viccuad, line 4: Nit: I would prefer for the disclosure to be at the bottom, it is important, but giving it too much prominence may send the message that we don't trust the project state, which isn't true. (Also, I would prefer it to be named "Security disclosure").

viccuad:
I wonder if applying suggestions from https://github.com/kubewarden/docs/pull/163#issuecomment-1383762038 or https://github.com/kubewarden/docs/pull/163#discussion_r1085266912 changes URI of content.

If so, I would like to lint https://github.com/kubewarden/kubewarden.io/tree/main/content/blog with a deadlink tool, and open a PR to update the blog links together with this one.

jvanz:
There is a duplicate menu item
image

sidebars.js
nunix: Applied suggestion

nunix:
> I wonder if applying suggestions from [#163 (comment)](https://github.com/kubewarden/docs/pull/163#issuecomment-1383762038) or [#163 (comment)](https://github.com/kubewarden/docs/pull/163#discussion_r1085266912) changes URI of content.
> 
> If so, I would like to lint https://github.com/kubewarden/kubewarden.io/tree/main/content/blog with a deadlink tool, and open a PR to update the blog links together with this one.

The links remain the same as the pages stay in the same filesystem path which the link is created from.
The Sidebar is more a virtual tree which doesn't impact the links. So no actions needed on the blog posts.

jvanz (approved):

flavio (approved):
Awesome, let's merge it!

PR #167, opened 2023-03-01
Fix missing favicon

kravciak:
Used the same icon as kubewarden.io


nunix (approved):

nunix:
Good catch, thank you for raising the PR @kravciak

viccuad:
TIL that one can and should use svg favicons!

PR #177, opened 2023-03-20
chore(deps): update docusaur

flavio:
Use latest version of docusaur

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,9 @@
     "write-heading-ids": "docusaurus write-heading-ids"
   },
   "dependencies": {
-    "@docusaurus/core": "^2.2.0",
-    "@docusaurus/preset-classic": "^2.2.0",
-    "@docusaurus/theme-search-algolia": "^2.2.0",
+    "@docusaurus/core": "^2.3.1",
+    "@docusaurus/preset-classic": "^2.3.1",
+    "@docusaurus/theme-search-algolia": "^2.3.1",
     "@mdx-js/react": "^1.6.22",
     "clsx": "^1.1.1",
     "prism-react-renderer": "^1.3.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2349,10 +2349,10 @@
     "@docsearch/css" "3.1.1"
     algoliasearch "^4.0.0"
 
-"@docusaurus/core@2.2.0", "@docusaurus/core@^2.2.0":
-  version "2.2.0"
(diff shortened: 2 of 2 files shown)


yarn.lock
```diff
@@ -2349,7 +2349,7 @@
     "@docsearch/css" "3.1.1"
     algoliasearch "^4.0.0"
 
-"@docusaurus/core@2.2.0", "@docusaurus/core@^2.2.0":
+"@docusaurus/core@2.2.0":
```
viccuad, line 2352: I would have expected 2.2.0 and its dependencies to get dropped. Is that ok?

jvanz (approved):

yarn.lock
jvanz: Good question. I think yarn should does that. @jordojordo do you know if this is okay?

btat: It's possible `"@docusaurus/theme-search-algolia": "^2.2.0"` not being bumped is causing this.

package.json
```diff
@@ -14,8 +14,8 @@
     "write-heading-ids": "docusaurus write-heading-ids"
   },
   "dependencies": {
-    "@docusaurus/core": "^2.2.0",
-    "@docusaurus/preset-classic": "^2.2.0",
+    "@docusaurus/core": "^2.3.1",
+    "@docusaurus/preset-classic": "^2.3.1",
     "@docusaurus/theme-search-algolia": "^2.2.0",
```
btat, line 19: I think this needs to be bumped as well. The [docs](https://docusaurus.io/docs/installation#updating-your-docusaurus-version) mention 

> Note that all `@docusaurus/`-namespaced packages should be using the same version.

yarn.lock
jordojordo: @btat is right, running `yarn list @docusaurus/core` yields:
```console
$ yarn list @docusaurus/core
yarn list v1.22.19
├─ @docusaurus/core@2.3.1
└─ @docusaurus/theme-search-algolia@2.2.0
   └─ @docusaurus/core@2.2.0
```

package.json
flavio: Good catch, this should be fixed now

yarn.lock
flavio: I fixed that

flavio:
Thanks for the reviews, please take another look to ensure everything is fine.

viccuad (approved):
LGTM!

PR #182, opened 2023-04-19
fix: host capabilities specifications

### LINKED ISSUE #181: Check if reference documentation matches of what is send over the wire
Double check if the reference documentation of the host capabilities function available matches of what is send over the wire in the guest and host communication. If there is some difference, fix the docs.

KhaledEmaraDev:
## Description
Fix Host Capabilities Specifications

Fix #181

### DIFF
--- a/docs/writing-policies/spec/host-capabilities/02-signature-verifier-policies.md
+++ b/docs/writing-policies/spec/host-capabilities/02-signature-verifier-policies.md
@@ -282,16 +282,22 @@ policy that verifies signatures can use:
   # mandatory: image URI to verify
   "image": string,
   # PEM-encoded certificated used to
-  # verify the signature
-  "certificate": string,
+  # verify the signature.
+  # The certificate is UTF-8 encoded.
+  # It's an array of bytes of the unicode code pointers of a PEM encoded
+  # certificate string.
+  "certificate": [byte(int), ..., byte(int)],
   # Optional - certificate chain used to
   # verify the provided certificate.
   # When not specified, the certificate
-  # is assumed to be trusted
+  # is assumed to be trusted.
+  # The certificate is UTF-8 encoded.
+  # It's an array of bytes of the unicode code pointers of a PEM encoded
+  # certificate string.
   "certificate_chain": [
-    string,
+    [byte(int), ..., byte(int)],
     ...
-    string
+    [byte(int), ..., byte(int)]
   ], 
   # Require the signature layer to have
   # a Rekor bundle.
@@ -346,20 +352,22 @@ Marked for deprecation:
 
 ```hcl
 {
-  # **mandatory**: image URI to verify
-  "image": string,
-  "pub_keys": [
-    # PEM-encoded public keys
-    string
+  "SigstorePubKeyVerify": {
+    # **mandatory**: image URI to verify
+    "image": string,
+    "pub_keys": [
+      # PEM-encoded public keys
+      string
     ],
-  # optional:
-  "annotations": [
+    # optional:
+    "annotations": [
       # signature annotations
       {
         "key": string,
         "value": string
       },
     ]
+  }
 }
 ```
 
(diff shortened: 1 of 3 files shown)


jvanz (changes requested):

docs/writing-policies/spec/host-capabilities/02-signature-verifier-policies.md
```diff
@@ -282,16 +282,20 @@ policy that verifies signatures can use:
   # mandatory: image URI to verify
   "image": string,
   # PEM-encoded certificated used to
-  # verify the signature
-  "certificate": string,
+  # verify the signature.
+  # The certificate is UTF-8 encoded.
+  # Each byte is a UTF-8 rune.
+  "certificate": [byte(int), ..., byte(int)],
```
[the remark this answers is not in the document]
viccuad: I agree with this.

flavio, line 288: I agree with the others, what about something like that:

```suggestion
  # Certificated used to
  # verify the signature.
  # Array of bytes containing the PEM
  # encoded certificate.
  "certificate": [byte],
```

docs/writing-policies/spec/host-capabilities/02-signature-verifier-policies.md
```diff
@@ -282,16 +282,20 @@ policy that verifies signatures can use:
...
-  # is assumed to be trusted
+  # is assumed to be trusted.
+  # The certificate is UTF-8 encoded.
+  # Each byte is a UTF-8 rune.
   "certificate_chain": [
-    string,
+    [byte(int), ..., byte(int)],
     ...
-    string
+    [byte(int), ..., byte(int)]
```
flavio, line 298: Same suggestion applies here

docs/writing-policies/spec/host-capabilities/05-crypto.md
```diff
@@ -28,18 +28,29 @@ performing cryptographic checks exposed by the host:
 <td>
 
 ```hcl
+# Certificate:
+{
+  # **mandatory**: Which encoding is used by the certificate
+  # Either the string "Pem" or "Der"; case sensetive.
```
flavio, line 34: Is this really case sensitive? I think we should change the policy-evaluator code to make it case insensitive.
If you agree with that, let's open an issue to track that

docs/writing-policies/spec/host-capabilities/05-crypto.md
```diff
@@ -28,18 +28,29 @@ performing cryptographic checks exposed by the host:
...
+  "encoding": string,
+  # Actual certificate
+  # The certificate is UTF-8 encoded.
+  # Each byte is a UTF-8 rune.
+  "data": [byte(int), ..., byte(int)]
```
flavio, line 39: I would use an adaptation of the suggestion I made earlier

docs/writing-policies/spec/host-capabilities/05-crypto.md
```diff
@@ -28,18 +28,29 @@ performing cryptographic checks exposed by the host:
...
+}
+
 {
   # **mandatory**: PEM-encoded certificate to verify
-  "certificate": string,
+  "cert": Certificate,
   # optional:
   "cert_chain": [
       # list of PEM-encoded certs, ordered by trust
       # usage (intermediates first, root last)
```
flavio, line 48: This is a list of Certificate objects, as defined before. Is that right? In that case we can drop the `PEM` string, because a `Certificate` object can be both PEM and DER

docs/writing-policies/spec/host-capabilities/05-crypto.md
KhaledEmaraDev: I agree. I created an issue and a PR to solve it.

Issue: https://github.com/kubewarden/policy-sdk-rust/issues/93
PR: https://github.com/kubewarden/policy-sdk-rust/pull/94

CC: @flavio @jvanz

viccuad: linked PR looks good to me, approving this.

viccuad (approved):

jvanz (approved):

PR #185, opened 2023-05-10
fix: telemetry is enabled only in the controller.

jvanz:
## Description

Updates the telemetry documentation to reflect the latest changes in the controller where is not necessary to enable telemetry in the policy servers. Now the user just need to enable telemetry on the controller and it will configure policy server to has telemetry as well.

Related to https://github.com/kubewarden/kubewarden-controller/pull/442


flavio (approved):
LGTM

viccuad (approved):

PR #197, opened 2023-07-05
Fixes Gatekeeper distribute example command

nnelas:
## Description

It seems that `kwctl manifest` is no longer a valid command and from what I could tell, we should use instead `kwctl scaffold manifest`

## Test

https://deploy-preview-197--silly-bunny-8cedd0.netlify.app/writing-policies/rego/gatekeeper/distribute#deploying-on-kubernetes


jvanz (approved):

jvanz:
@nnelas thank you for the contribution!

PR #209, opened 2023-08-15
Quick-start install doc review

jhkrug:
Minor changes to structure and language. 

Split commands into separate console blocks. So no multiline command copies from the console blocks. My preference is not to be pasting multiple lines of commands after I copy them.

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -1,20 +1,19 @@
 ---
-sidebar_label: "Quick Start"
-title: ""
+sidebar_label: "Quick start"
+title: "Quick start"
 ---
 
-# Quick Start
 
 The Kubewarden stack is made of the following components:
 
 * An arbitrary number of `ClusterAdmissionPolicy` resources: this is how policies are defined inside Kubernetes
-* An arbitrary number of `PolicyServer` resources: this component represents a Deployment of a Kubewarden `PolicyServer`. The policies defined by the administrators are loaded and evaluated by the Kubewarden `PolicyServer`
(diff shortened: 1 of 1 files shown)


docs/quick-start.md
```diff
@@ -1,84 +1,108 @@
...
 
-# Quick Start
 
-The Kubewarden stack is made of the following components:
+The Kubewarden stack comprises:
 
-* An arbitrary number of `ClusterAdmissionPolicy` resources: this is how policies are defined inside Kubernetes
-* An arbitrary number of `PolicyServer` resources: this component represents a Deployment of a Kubewarden `PolicyServer`. The policies defined by the administrators are loaded and evaluated by the Kubewarden `PolicyServer`
-* A Deployment of `kubewarden-controller`: this is the controller that monitors the `ClusterAdmissionPolicy` resources and interacts with the Kubewarden `PolicyServer` components
+* Some `ClusterAdmissionPolicy` resources: this is how policies are defined inside Kubernetes
```
jvanz, line 11: I think we can add a similar item for `AdmissionPolicy`

jvanz (approved):
LGTM. Just a small suggestion

docs/quick-start.md
jhkrug: Ok, I've done that. Does what I've added correctly describe the tech, the difference in talking about ClusterAdmissionPolicy and AdmissionPolicy? I don't want to be introducing early stage confusion :-). Thanks.

jhkrug: Maybe 'defined namespace' is better than 'named namespace' ...

PR #214, opened 2023-08-22
Refresh of language in Common tasks page

jhkrug:
and to reflect use of artifact hub instead of hub.kubewarden.io.


jhkrug:
There are a couple of TODO/FIXIT that I need to address in this.

docs/tasks.md
```diff
@@ -50,14 +50,12 @@ Depending on your role, `kwctl` will help you in the following non-exhaustive sc
 
 #### Installation
 
-`kwctl` binaries for the stable releases are directly available from the [GitHub repository](https://github.com/kubewarden/kwctl/releases).
+`kwctl` binaries for the stable releases are available from the [GitHub repository](https://github.com/kubewarden/kwctl/releases).
 
-> **NOTE**: If you want to build `kwctl` from the development branch, you need to install [Rust](https://www.rust-lang.org/tools/install). And for building `kwctl`, please refer to the `Build kwctl from source` section in the [GitHub repo](https://github.com/kubewarden/kwctl).
+> **NOTE**: If you want to build `kwctl` from the development branch, you need a [Rust](https://www.rust-lang.org/tools/install) development environment. To build `kwctl`, refer to the "Build kwctl from source" section in the [GitHub repo](https://github.com/kubewarden/kwctl).  <!--FIXME: Can't find this section.-->
```
jvanz, line 55: I think we can point to the repository and remove the reference to the section.

jvanz (commented):
LGTM. I have dropped some comments answering your questions about the text.

docs/tasks.md
```diff
@@ -234,9 +232,7 @@ After you have generated the `ClusterAdmissionPolicy` and applied it to your Kub
       mutating: false
     ```
   
-    > TIP: By default, the `name` value is set to `generated-policy`. You might want to edit it before you deploy the `ClusterAdmissionPolicy`. 
-    >
-    > NOTE: To avoid confusion, the value above has been set to `privileged-pods`.
+    > TIP: By default, the `name` value is set to `generated-policy`. You might want to edit it before you deploy the `ClusterAdmissionPolicy`. The name in the immediately previous example has been set to `privileged-pods`.
```
jvanz, line 235: Maybe we can replace this kind of `> TIP` by [docusaurus admonitions](https://docusaurus.io/docs/markdown-features/admonitions) as we do above in the page.

jhkrug: Admonitions are great and stand out well. The tip is a bit less important so it's probably ok to not have it standout quite as much. Both mechanisms should be used sparingly otherwise a reader stops paying attention to them. But, yeah, we have two different styles of 'admonitions' which is not great from a consistency point of view. Another iteration perhaps.

jhkrug: Hmmm, I think consistency is the best policy,so hold on a few minutes. :-)

PR #228, opened 2023-08-31
Refresh of context aware policies.

jhkrug:
There is a question in a TODO: at the end of the page. Thanks.

### DIFF
--- a/docs/explanations/context-aware-policies.md
+++ b/docs/explanations/context-aware-policies.md
@@ -1,30 +1,28 @@
 ---
-sidebar_label: "Context Aware Policies"
-title: ""
+sidebar_label: "Context aware policies"
+title: "Context aware policies"
+description: The description
+keywords: [kubewarden, context-aware]
 ---
 
-# Context Aware Policies
+Developers can create policies that fetch information from a Kubernetes cluster at run time.
+These are context aware policies.
+Context aware policies can determine whether an `AdmissionRequest` is acceptable using information from resources deployed in the cluster.
 
-Context aware policies allow policy authors to fetch information from the
-Kubernetes cluster at evaluation time. This means that context aware policies can determine
-whether a `AdmissionRequest` has to be accepted or rejected based on other resources already
-deployed in the cluster.
-For example, a policy could cross validate information among two different resources type.
-
-:::caution
-Context aware policies are available only in Kubewarden version v1.6.0 or greater.
+:::note
+Context aware policies are only available in Kubewarden versions ≥ v1.6.0.
 :::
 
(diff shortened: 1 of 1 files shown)


docs/explanations/context-aware-policies.md
```diff
@@ -54,27 +52,29 @@ spec:
...
+The same policy can't justify having access to `Secret` resources.
 
-When reviewing the list of resources a context aware policy wants to have access to, approach this process with the same mindset used when evaluating the privileges granted to a smart phone App you're about to install on your mobile device.
-Would you install a torch App that requires access to your address book or your location?
+Policies should have the least access needed to function correctly.
 
-While policies have read-only access to Kubernetes resources, a malicious attacker might abuse a Kubewarden policy to exfiltrate sensitive data from the cluster.
 :::
 
+<!--TODO: What is this note for? what does it mean? Is it relevant to context aware policies?-->
```
flavio, line 73: Yes, this note is relevant to the context aware policies.

Some background, all the Kubernetes resources are identified via `apiVersion` and `kind`. Usually `apiVersion` is a string following this pattern: `<api>/<version>`. For example, the Deployment objects have `apps/v1` as `apiVersion`, while Ingress have `networking.k8s.io/v1`. However, all the resources that belong to the `core` group (like Pod, Service, Node, ConfigMap, Secret,...) have a different pattern: `<version>`.

This is relevant for context aware policies because the admin must define the allow list of resources the policy can read. 

Having this rule would not work (meaning the policy would not be granted access to Pod):

```yaml  contextAwareResources:
    - apiVersion: "core/v1"
      kind: "pod"
```

While this one would:

```yaml  contextAwareResources:
    - apiVersion: "v1"
      kind: "pod"
```

BTW, most Kubernetes users should not be surprised by that syntax give this is what is used when defining a Pod, Service,... via a yaml file.

flavio (commented):
LGTM, I clarified the question @jhkrug asked.

docs/explanations/context-aware-policies.md
jhkrug: Thanks Flavio.

jhkrug: Hi @flavio I've made changes to the note at the end. Hopefully it is technical accurate, clear and readable. Thanks.

flavio (approved):
LGTM, it looks like this can be merged. Am I right?

viccuad (approved):
LGTM! I really like the refactor, I think it hits the sweet spot of being succint and to the point (I look for that on dev docs), and have all the content.

PR #237, opened 2023-09-14
docs: update telemetry guide

fabriziosestito:
## Description
Update the telemetry guide with the latest chart version, 

Add Kubewarden controller Prometheus configuration that was missing even if we were referring to metrics exported by the controller in the guide.


flavio (approved):
LGTM, thanks a lot for the fixes!

viccuad (approved):
Many thanks for the clarification and refreshing the chart versions, that is quite helpful.

PR #241, opened 2023-09-15
fix: policy reporter values for air gap environments.

### LINKED ISSUE #239: Update documentation mapping the values change when user want to enable policy reporter in a air gap environment

jvanz:
## Description

Updates the operator manual about air gap environments adding a text explaining the required additional values that must be provided when when the Policy Reporter subchart is installed together with the kubewarden-controller chart.

Fix #239


docs/operator-manual/airgap/02-install.md
```diff
@@ -121,6 +121,28 @@ helm install --wait -n kubewarden \
   --set global.cattle.systemDefaultRegistry=<REGISTRY.YOURDOMAIN.COM:PORT>
 ```
 
+:::caution
+If you want to use the Policy Reported subchart available in the
+`kubewarden-controller` chart you need to define other values specific for the
+subchart in order to work in a air gap environment. See an example below:
```
fabriziosestito, line 127: maybe: `in an air-gapped environment`

viccuad (approved):
LGTM modulo Fabrizio's suggestion.

flavio (approved):
LGTM, once fabrizio's feedback is taken into account

PR #245, opened 2023-09-21
Revert to default fonts.

jhkrug:
The default fonts work better together, especially the code font.


jhkrug:
Side by side, default on left, current on right.

image

flavio:
thanks! this is a nice improvement

PR #261, opened 2023-10-06
fix: Remove duplicated entry for context-aware-policies

viccuad:
## Test

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -116,7 +116,6 @@ module.exports = {
         {
           Comparisons: ["explanations/opa-comparison"],
         },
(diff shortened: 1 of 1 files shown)


sidebars.js
```diff
@@ -116,7 +116,6 @@ module.exports = {
         {
           Comparisons: ["explanations/opa-comparison"],
         },
-        "explanations/context-aware-policies",
```
viccuad, line 119: it's in line 115 too, see a bit above.

jhkrug (approved):
Great!

PR #262, opened 2023-10-09
Update tracing and metrics installation snippets

flavio:
Be explicit about the version to be used, this ensures reproducibility.

### DIFF
--- a/docs/operator-manual/telemetry/metrics/01-quickstart.md
+++ b/docs/operator-manual/telemetry/metrics/01-quickstart.md
@@ -82,6 +82,7 @@ At time of writing the latest chart version is `51.0.2`
 ```console
 helm install --wait --create-namespace \
   --namespace prometheus \
+  --version 51.0.2 \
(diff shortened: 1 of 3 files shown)


jhkrug:
Hi, we should discuss versioning in the docs briefly in a parking lot. Thanks.

Yes, reproducibility, that's great if needed. If needed, then can we do auto updates with the likes of updatecli to make sure that nothing gets missed? If not, then some sort of checklist for all hard coded version number in the docs might be a good idea.

Is being more general possible? Saying use the latest version? Or something else generic.

There are always going to be exceptions, but we can discuss the general approach. 

This conversation is also taking place in Epinio.

jhkrug (commented):
Not quite doing this right. Sorry.

I have made some comments on the Conversation page.

viccuad (approved):
Approving.
These specific versions are needed because opentelemetry is still a moving target and they break compatibility between versions still.

fabriziosestito (approved):
LGTM

PR #263, opened 2023-10-09
CRD docs integration

jhkrug:
Do you like the look of this? I've just added the generated file contents into the page to see what it looks like.

@flavio , @viccuad , @jvanz


viccuad:
@jhkrug  it's starting to look good, but the `matchPolicy` file entry breaks the rest of the table.
Compare with: https://doc.crds.dev/github.com/kubewarden/kubewarden-controller/policies.kubewarden.io/ClusterAdmissionPolicy/v1@v1.7.1

viccuad:
This is personal preference, but I wonder if we could not generate the docs for old `v1alpha` versions, and only show `v1`, as happens in https://doc.crds.dev/github.com/kubewarden/kubewarden-controller/policies.kubewarden.io/AdmissionPolicy/v1@v1.7.1 (see top right).

jhkrug:
> @jhkrug it's starting to look good, but the `matchPolicy` file entry breaks the rest of the table. Compare with: https://doc.crds.dev/github.com/kubewarden/kubewarden-controller/policies.kubewarden.io/ClusterAdmissionPolicy/v1@v1.7.1#spec-matchPolicy

This because the documentation source contains markdown for a list. A markdown list is not permitted in a markdown table cell. At least not in docusaurus. So, you can't put markdown formatting in the crd source. It's likely not to work. It works for asciidoc. Which we don't use, sadly perhaps.

viccuad:
Maybe we can generate the crd docs in markdown instead, that could be a possibility too.

jhkrug:
> Maybe we can generate the crd docs in markdown instead, that could be a possibility too.

That's what I did for the demo.

jhkrug:
I can take a look at the docs source file for this and try opening a PR in there.

jhkrug:
> > @jhkrug it's starting to look good, but the `matchPolicy` file entry breaks the rest of the table. Compare with: https://doc.crds.dev/github.com/kubewarden/kubewarden-controller/policies.kubewarden.io/ClusterAdmissionPolicy/v1@v1.7.1#spec-matchPolicy
> 
> This because the documentation source contains markdown for a list. A markdown list is not permitted in a markdown table cell. At least not in docusaurus. So, you can't put markdown formatting in the crd source. Plain, vanilla, unformatted text only. It's likely not to work. It works for asciidoc. Which we don't use, sadly perhaps.

You can use a html list in policy_types.go and it works as well. So, that's another option.

jhkrug:
So, it appears that the most you can get away with regarding docstring formatting is a html encoded list and list items. And I'd discourage that as well. Too prone to error and complications. 

Some of the docstrings have code blocks in them. They don't work. I'd suggest an iteration of the docstrings. There is perhaps too much material in there. Keep it as short as possible, with links to doc pages as necessary for further explanations.

So why does it currently work? Well, different tool chains, but it doesn't really.

image

viccuad (approved):
LGTM!
Super happy with this! This section was also the last one not accesible offline, now we can ship the docs together with the helm-charts, make them available even on airgap, or provide them in the UI!

flavio:
This looks really good to me. 

What about adding a more explicit link to that, right now this is kinda buried under sub-menus on the left sidebar.

I propose to do something like that:

image


If you're fine with that I can add this small change to this branch

viccuad:
I'm ok with making the CRDs more explicit, maybe just under "Reference docs" and not "Reference docs -> Operator Manual".
I'm also ok linking to the CRDs docs from the CRD names in the quickstart (I don't understand much the shared screenshot).

flavio:
> I'm also ok linking to the CRDs docs from the CRD names in the quickstart (I don't understand much the shared screenshot).

Forgive me, I wasn't really explicit about the change I made. I've just added the new "TIP" section, the one in green.

flavio (approved):
thanks for the new changes!

Issue #234, opened 2023-09-11
Another look at Audit Scanner


jhkrug:
Low priority. Just another pass through.

PR #264, opened 2023-10-10
Review and refresh of audit scanner overview documents.

jhkrug:
Will complete #234.

I never got round to doing this properly a couple of months ago. Apologies.


docs/explanations/audit-scanner/audit-scanner.md
```diff
@@ -1,89 +1,78 @@
...
+Some days later, Alice, the Kubernetes administrator, enforces a Kubewarden policy that prohibits the creation of privileged containers.
+The Pod deployed by Bob keeps running in the cluster as it already exists.
+
+A report generated by the audit scanner lets Alice identify all the workloads that are violating creation policies.
+This includes the WordPress Pod created by Bob.
+
+The audit scanner operates by:
+
+- identifying all the resources to audit
+- for each it builds a fake admission request with the resource's data
```
viccuad, line 39:
```suggestion
- for each it builds a synthetic admission request with the resource's data
```

I went with "synthetic" in the blog post because of style, I wonder if it would be a good fit here too. The only appearance of the word "fake" is here.

jhkrug: Thanks!

docs/explanations/audit-scanner/limitations.md
```diff
@@ -51,7 +52,10 @@ spec:
         - UPDATE
 ```
 
-The `apiGroups`, `apiVersions` and `resources` attributes can use the special `*` inside of them. This is a wildcard
-symbol that causes the policy to match all the values used inside of the field.
+<!--TODO:Is this OK, to me it looks like we are saying how to use a
+wildcard. Then we say that wildcards are ignored. Is that right?-->
```
flavio, line 56: We are pointing out that it's fine to use wildcards for certain fields of a policy CR. However, the user must be aware that these type of policies are not going to be evaluated by the audit checks.

docs/explanations/audit-scanner/policy-reports.md
```diff
@@ -1,75 +1,74 @@
...
+[policy-reporter](https://kyverno.github.io/policy-reporter)
+open source project for monitoring and observability of the PolicyReport CRDs.
 
 ## Policy Reporter UI
 
-The Policy Reporter is shipped as a subchart of `kubewarden-controller`, refer
-to the [Audit Scanner Installation](../../howtos/audit-scanner) page for more
-info.
+The Policy Reporter is shipped as a subchart of `kubewarden-controller`.
+Se the [Audit Scanner Installation](../../howtos/audit-scanner)
```
flavio, line 47:
```suggestion
To the [Audit Scanner Installation](../../howtos/audit-scanner)
```

flavio (commented):
Looks way better.

I've replied to your request for comments and left a minor suggestion.

flavio (approved):

PR #307, opened 2023-11-14
chore: Point to artifacthub.io instead of hub.kubewarden.io

viccuad:
## Test


jhkrug (commented):
OK, but are you using tools that add unnecessary line breaks, into keywords, and unnecessary spaces into tables?

viccuad:
@jhkrug  argh, indeed. I'm using https://prettier.io from my editor, as we use it for the source code repos, and committed the whole file, instead of the line I changed, sorry.
Repushed just with the minimal change.

jhkrug:
> @jhkrug argh, indeed. I'm using https://prettier.io from my editor, as we use it for the source code repos, and committed the whole file, instead of the line I changed, sorry. Repushed just with the minimal change.

It's ok, I was just wondering. Maybe it should be standard.

PR #315, opened 2023-11-28
Update grafana dashboard screenshot

viccuad:
## Description

Fix https://github.com/kubewarden/policy-server/issues/597


docs/operator-manual/telemetry/metrics/01-quickstart.md
```diff
@@ -205,10 +205,7 @@ Another option is import it directly from the Grafana.com website. For this:
 
 You should be able to see the dashboard similar to this:
 
-![Dashboard 1](/img/grafana_dashboard_1.png)
-![Dashboard 2](/img/grafana_dashboard_2.png)
-![Dashboard 3](/img/grafana_dashboard_3.png)
-![Dashboard 4](/img/grafana_dashboard_4.png)
+![Dashboard](/img/grafana_dashboard.png)
```
viccuad, line 208: It seems we are using assets that are global and not versioned, so I haven't removed the `grafana_dashboard_X.png` files from `docs/images`. Hence they are present in `docs/images`, `static/image`, and of course on the versions (`version-1.9/images/` for example).

See:
https://docusaurus.io/docs/versioning#global-or-versioned-collocated-assets

jhkrug: I think that's ok. Seems to be similar to Epinio. No versioned assets

jhkrug (approved):
ok

PR #319, opened 2023-11-30
Docusaurus V3 upgrade.

jhkrug:


flavio (approved):
It builds now :)

The preview website looks good

viccuad (commented):
Thanks for the porting! Preview looks fine, also the mermaid diagram in upgrade path.

viccuad (approved):

jhkrug:
All straightforward. The problems with escaping '{' and '}' had been noted by Nuno previously. There were a few of those to catch in the versioned docs. All in a file that is  generated for the API in the kubewarden-controller repository. That was fixed in https://github.com/kubewarden/kubewarden-controller/pull/588.

Issue #326, opened 2023-12-29
Review and update frontmatter across all docs markdown files


jhkrug:
Check format. Make sure title, description, keywords, doc-persona, doc-type, doc-topic are all defined for each file.

This is to set up for future coverage tooling.

PR #338, opened 2024-01-23
Miscellaneous front matter updates

jhkrug:
Will complete #326


jhkrug:
Converted to a draft while I make some further additions.

flavio:
@jhkrug should we merge it? seems good to me

jhkrug:
> @jhkrug should we merge it? seems good to me

I think so. It's a largish change so I'll do it in the morning and keep an eye on it. Just to make sure.

PR #347, opened 2024-01-29
feat: new OCI image index capability.

### LINKED ISSUE #346: Protocol specification for the capability to allow users to fetch OCI manifests
Update the docs section about the protocol specification to document the new capability that allows users to fetch OCI manifests. 

### Acceptance criteria

- Describe the expected behavior of the capability
- Describe the expected input data 
- Describe the expected output data

jvanz:
## Description

Adds new section in the container registry capabilities showing how users can fetch OCI image index to get information about OCI manifests in different platforms.

Fix #346

### DIFF
--- a/docs/writing-policies/spec/host-capabilities/03-container-registry.md
+++ b/docs/writing-policies/spec/host-capabilities/03-container-registry.md
@@ -79,3 +79,154 @@ the payload would be the following ones:
 * Input payload: `"busybox:latest"`
 * Output payload: `{ "digest": "sha256:69e70a79f2d41ab5d637de98c1e0b055206ba40a8145e7bddb55ccc04e13cf8f"}`
 
+## OCI Image manifest
+
+This function fetches OCI images manifest. When available, this information can
+be used to identify specific images manifests, for one or more platforms. Or a
+single image manifest for the image.
+
+### Caching
+
+Computing the digest involves a series of network requests between the
+Kubewarden policy host and the remote registry. These operations can be time
+expensive so the results are cached for 1 minute.
+
+### Authentication
+
+Interactions with private registries require the Kubewarden policy host to
+authenticate against the remote registry.
+
+The policy host uses the same set of credentials as that used to fetch policies
+from the remote registry.
+
+### Communication protocol
+
+This is the description of the waPC protocol used to expose this capability:
+
+<table>
+<tr>
+<th> waPC function name </th> <th> Input payload </th> <th> Output payload </th>
+</tr>
+
+<tr>
+<td>
+
+`v1/image_manifest`
+
+</td>
+<td>
+
+```hcl
+# OCI URI - JSON encoded string
+string
+```
+
+</td>
+
+<td>
+
+```hcl
+{
+    "schemaVersion": 2,
+        "mediaType": "application/vnd.oci.image.index.v1+json",
+        "manifests": [
+        {
+            "mediaType": "application/vnd.oci.image.manifest.v1+json",
+            "digest": "sha256:7d5e84b9314ba7058bfa209881919146ffc4a89c5ba14cfa8270a18c8d418c44",
+            "size": 1048,
+            "platform": {
+                "architecture": "amd64",
+                "os": "linux"
+            }
+        },
+        {
+            "mediaType": "application/vnd.oci.image.manifest.v1+json",
+            "digest": "sha256:ae1e98105555f3983496c15c70dafc87639b77830953d6470694b47e0e063e25",
+            "size": 1048,
+            "platform": {
+                "architecture": "arm64",
+                "os": "linux"
+            }
+        },
+        {
+            "mediaType": "application/vnd.oci.image.manifest.v1+json",
+            "digest": "sha256:e9fc0f86e26366abf29fd29e8c09d00df717f9038fd4298eaa39a0a2b4361fa4",
+            "size": 566,
+            "annotations": {
+                "vnd.docker.reference.digest": "sha256:7d5e84b9314ba7058bfa209881919146ffc4a89c5ba14cfa8270a18c8d418c44",
+                "vnd.docker.reference.type": "attestation-manifest"
+            },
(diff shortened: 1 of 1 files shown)


docs/writing-policies/spec/host-capabilities/03-container-registry.md
```diff
@@ -79,3 +79,113 @@ the payload would be the following ones:
 * Input payload: `"busybox:latest"`
 * Output payload: `{ "digest": "sha256:69e70a79f2d41ab5d637de98c1e0b055206ba40a8145e7bddb55ccc04e13cf8f"}`
 
+## OCI Image index
```
flavio, line 82: I would call it `Image manifest`. The Image index is a [type of manifest](https://github.com/opencontainers/image-spec/blob/main/image-index.md):

> The image index is a higher-level manifest which points to specific [image manifests](https://github.com/opencontainers/image-spec/blob/main/manifest.md), ideal for one or more platforms. While the use of an image index is OPTIONAL for image providers, image consumers SHOULD be prepared to process them.

There's no API to fetch the Image index from a registry, that's because you can obtain it by fetching the image manifest.

Because of that, I propose to replace all the `index` instances inside of this PR with `manifest`

docs/writing-policies/spec/host-capabilities/03-container-registry.md
```diff
@@ -79,3 +79,113 @@ the payload would be the following ones:
...
+</tr>
+
+</table>
+
+For example, when requesting the manifest digest of the `ghcr.io/kubewarden/policy-server:v1.10.0` image,
+the payload would be the following ones:
+
+* Input payload: `"ghcr.io/kubewarden/policy-server:v1.10.0"`
+* Output payload: The same of the data used in the table above. Which follows
+the [OCI Image index specification](https://github.com/opencontainers/image-spec/blob/main/image-index.md)
```
flavio, line 190: nitpick, the result depends on what the remote image has. It could be a simple manifest, an image index or even a legacy manifest.

I would say something like:

```suggestion
Output payload: the body of the successful response obtained from the registry
```

flavio (changes requested):
The protocol spec is fine, just some minor fixes

fabriziosestito (approved):
LGTM when comments are addressed

docs/writing-policies/spec/host-capabilities/03-container-registry.md
```diff
@@ -79,3 +79,113 @@ the payload would be the following ones:
 * Input payload: `"busybox:latest"`
 * Output payload: `{ "digest": "sha256:69e70a79f2d41ab5d637de98c1e0b055206ba40a8145e7bddb55ccc04e13cf8f"}`
 
+## OCI Image index
+
+This function fetches OCI images index manifest. This information can
+be used to identify specific images manifests, for one or more platforms.
```
jhkrug, line 85:
```suggestion
This function fetches OCI image index manifests. This information can
be used to identify specific image manifests, for one or more platforms.
```

Just a suggestion, I hope I'm not changing the meaning, but I think it reads a little better.

docs/writing-policies/spec/host-capabilities/03-container-registry.md
```diff
@@ -79,3 +79,113 @@ the payload would be the following ones:
...
+
+### Caching
+
+Computing the digest involves a series of network requests between the Kubewarden
+policy host and the remote registry. These operation can be time expensive,
+because of that the results are going to be cached for 1 minute.
```
jhkrug, line 91:
```suggestion
Computing the digest involves a series of network requests between the Kubewarden
policy host and the remote registry. These operations can be time expensive so the results are cached for 1 minute.
```

docs/writing-policies/spec/host-capabilities/03-container-registry.md
```diff
@@ -79,3 +79,113 @@ the payload would be the following ones:
...
+
+### Authentication
+
+Interactions with private registries require the Kubewarden policy host to
+authenticate against the remote registry.
+
+The policy host will use the same set of credentials used to fetch policies
+from the remote registry.
```
jhkrug, line 99:
```suggestion
The policy host uses the same set of credentials as that used to fetch policies
from the remote registry.
```

jhkrug (commented):
Some minor suggestions.

docs/writing-policies/spec/host-capabilities/03-container-registry.md
jvanz: Maybe we can change the docs to call it `Image manifest` but making clear that the output can change depending of the OCI artifact manifest requested:

```console
$ crane manifest ghcr.io/kubewarden/policies/pod-privileged:latest | jq ""                                                                                                             
{
  "config": {
    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
    "mediaType": "application/vnd.wasm.config.v1+json",
    "size": 2
  },
  "layers": [
    {
      "digest": "sha256:6d803e1199b0c884741a43c6fbc2f1586839429f00681a3098116437d4210b71",
      "mediaType": "application/vnd.wasm.content.layer.v1+wasm",
      "size": 1285267
    }
  ],
  "schemaVersion": 2
}
```

Therefore, the output can be a `Image index` or a single `Image manifest`. And the policy must deal with it.

jvanz: @flavio  I've updated the docs changing the name and adding an statement that depending of the image the response can change. It can be a index manifest or a single image manifest.

flavio (approved):
LGTM, I think we can even merge it into the main branch. This is going to stay on the "Next" version of the docs, which is fine

docs/writing-policies/spec/host-capabilities/03-container-registry.md
viccuad: I"m happy with this approach, approving.

PR #350, opened 2024-02-08
Bump docusaurus 3.0.1 to 3.1.1.

jhkrug:
New DS picks up many invalid anchors that needed fixing. 

Those remaining need a closer look at the CRD doc generator.


viccuad (approved):
LGTM

PR #353, opened 2024-02-09
New docs layout

jhkrug:
First commit new org, will not build yet, broken links.
A backup before the weekend.


flavio:
I like it, some observations:

* The "howtos/verification configuration" could probably be moved under "howtos/secure supply chain"
* "reference/CRD" should probably be the 1st item of the "reference" sidebar menu

jhkrug:
> I like it, some observations:
> 
> * The "howtos/verification configuration" could probably be moved under "howtos/secure supply chain"
> * "reference/CRD" should probably be the 1st item of the "reference" sidebar menu

After thinking a little ---- Maybe CRDs should be seconds after 'Dependencies'?

docs/howtos/verification-config.md
```diff
@@ -1,10 +1,11 @@
 ---
 sidebar_label: Verification configuration
+sidebar_position: 45
 title: Verification configuration
 description: Verification configuration for Kubewarden.
 keywords: [kubewarden, kubernetes, verification configuration]
 doc-persona: [kubewarden-operator, kubewarden-integrator]
-doc-type: [tutorial, howto, reference]
+doc-type: [howto]
```
viccuad, line 8: I think this Verification configuration belongs under "reference". For this "verification configuration", I would:
- Swap the order of the contents in the file, first putting the "format", then the "example".
- Add a starting phrase stating what is it for: 
  > this `verification-config` format is used by `policy-server` to verify policy modules provenance, and
  > by `verify-image-signatures` policy to verify cluster images provenance. See ./howtos/secure-supply-chain for more info.
- Move the subsection of https://deploy-preview-353--silly-bunny-8cedd0.netlify.app/next/howtos/secure-supply-chain#signature-configuration-reference, which is a reference, inside the "Verification Configuration" page.
- Potentially rename from "Verification configuration" to "verification-config format".

docs/howtos/policy-servers/01-custom-cas.md
```diff
@@ -4,7 +4,7 @@ title: Using custom certificate authorities
 description: Using custom certificate authorities with Kubewarden policy servers.
 keywords: [kubewarden, kubernetes, custom certificate authorities]
 doc-persona: [kubewarden-operator, kubewarden-integrator]
-doc-type: [howto, reference]
+doc-type: [howto]
```
viccuad, line 7: Both this page `docs/howtos/policy-servers/01-custom-cas.md` and `howtos/custom-certificate-authorities.md` explain the `sources.yaml` config file for the policy-server. I would leave an example of usage and merge both `sources.yaml` config file explanations under a reference page.

docs/howtos/telemetry/metrics/_category_.json
```diff
@@ -0,0 +1,5 @@
+{
```
viccuad, line 1: I wonder if for opentelemetry, tracing, and metrics we need collapsed subsections that are empty. I would be ok showing in the sidebar the following:
> - Telemetry
>   * Opentelemetry quickstart
>   * Tracing quickstart
>   * Metrics quickstart


I don't expect each of the OTLM, Tracing, Metrics to grow; new production deployments will likely be downstream in the UI and automated, and will take a while to get community submissions to add to the docs.

viccuad:
> > I like it, some observations:
> > 
> > * The "howtos/verification configuration" could probably be moved under "howtos/secure supply chain"
> > * "reference/CRD" should probably be the 1st item of the "reference" sidebar menu
> 
> After thinking a little ---- Maybe CRDs should be second after 'Dependencies'?

I concur with having CRDs as 1st item. The policy user persona (most important) deals with them a lot when configuring policies. And also the dev persona. Plus, the dependencies one it's only looked by 1 persona and still seldomly.

jhkrug:
> > > I like it, some observations:
> > > 
> > > * The "howtos/verification configuration" could probably be moved under "howtos/secure supply chain"
> > > * "reference/CRD" should probably be the 1st item of the "reference" sidebar menu
> > 
> > 
> > After thinking a little ---- Maybe CRDs should be second after 'Dependencies'?
> 
> I concur with having CRDs as 1st item. The policy user persona (most important) deals with them a lot when configuring policies. And also the dev persona. Plus, the dependencies one it's only looked by 1 persona and still seldomly.

Ok, that's already done then.

docs/howtos/telemetry/metrics/_category_.json
jhkrug: Ok, I've done that for the next commit.

docs/howtos/verification-config.md
jhkrug: OK, that's in the next commit. Thanks.

docs/howtos/policy-servers/01-custom-cas.md
jhkrug: OK. I will need some more input. I expect I'm not understanding it but if they are referencing the same format `sources.yaml` why do the keys look different? `insecure_sources` vs `insecureSources`. Thanks

viccuad: You are right, they are indeed 2 different things.

There are:
1. The "`sources.yaml` file format", which is used by both `kwctl` and the `policy-server` binaries. This `sources.yaml` file format should go into the references page. It is the one at "distributing-policies/custom-certificate-authorities".
2. The PolicyServer `spec.sourceAuthorities` and `spec.insecureSources`. They are part of the the CRD definition, but too big to be part of the CRD reference page. IMHO  `spec.sourceAuthorities` and `spec.insecureSources` both should be part of a new reference page. The CRDs should be updated to include links from spec.sourceAuthorities and spec.insecureSources into that new reference page for them. They are closely related to the "`sources.yaml` file format", in that the controller reads the PolicyServer spec.sourceAuthorities and spec.insecureSources and creates a sources.yaml for it, when instantiating its deployment.

jhkrug: Hmm, is it acceptable to create a new card to do this.? Independent of the rest of the restructuring. To be done next. Not left languishing. :smiley:

viccuad: Indeed, sorry! Let's move this forward and iterate. Approving.

viccuad (approved):

docs/howtos/policy-servers/01-custom-cas.md
jhkrug: Ok. New issue in https://github.com/kubewarden/docs/issues/354

jvanz (approved):

PR #362, opened 2024-02-22
docs: add missing parameter

90er:
## Description

Add missing parameter if the namespaces doesn't already exist

## Test

(Docs)


jhkrug:
Thanks Wolfgang! Much appreciated. And just reaching out. Is your organization actively using Kubewarden? If so, maybe you can add yourselves to the ADOPTERS.md file. Thanks.

PR #367, opened 2024-03-04
source.yaml reference and reorganization.

### LINKED ISSUE #354: Rework Custom CAs docs
Following on from #353 comments by @viccuad. https://github.com/kubewarden/docs/pull/353#discussion_r1487368611

## Acceptance criteria

- [ ] Will be closed by #367.

## Context

@viccuad. Both this page `docs/howtos/policy-servers/01-custom-cas.md` and `howtos/custom-certificate-authorities.md` explain the `sources.yaml` config file for the policy-server. I would leave an example of usage and merge both `sources.yaml` config file explanations under a reference page.

@jhkrug. OK. I will need some more input. I expect I'm not understanding it but if they are referencing the same format `sources.yaml` why do the keys look different? `insecure_sources` vs `insecureSources`. Thanks.

@viccuad. You are right, they are indeed 2 different things.

There are:
1. The "`sources.yaml` file format", which is used by both `kwctl` and the `policy-server` binaries. This `sources.yaml` file format should go into a  "`sources.yaml` file format" reference page. Right now it is at "distributing-policies/custom-certificate-authorities".
2. The "PolicyServer `spec.sourceAuthorities` and `spec.insecureSources` format". They are part of the the CRD definition, but too big to be part of the CRD reference page. IMHO  `spec.sourceAuthorities` and `spec.insecureSources` both should be part of a new reference page. The CRDs should be updated to include links from spec.sourceAuthorities and spec.insecureSources into that new reference page for them. They are closely related to the "`sources.yaml` file format", in that the controller reads the PolicyServer spec.sourceAuthorities and spec.insecureSources and creates a sources.yaml for it, when instantiating its deployment.

jhkrug:
Initial commit for docs:#354. WIP.

Will close #354 when completed.

I need to discuss further with @viccuad. I'm not quite understanding the requirement in the second part of #354.


jhkrug:
Waiting for @viccuad to return.

docs/reference/sources_yaml.md
```diff
@@ -0,0 +1,118 @@
...
+    "source_authorities": {
+        "host.k3d.internal:5000": [
+            {"type": "Data","data":"pem cert 1"},
+            {"type": "Data","data":"pem cert 2"}
+        ]
+    }
+}
+```
+
+### Insecure sources
```
viccuad, line 67:
```suggestion
### Insecure sources section
```

docs/reference/sources_yaml.md
```diff
@@ -0,0 +1,118 @@
...
+:::note
+
+It's usually fine to use `insecure_sources` when using local registries or
+HTTP servers for development.
+It avoids the burden of managing certificates.
+Clearly, it's not for production use.
+
+:::
+
+### Source authorities
```
viccuad, line 92:
```suggestion
### Source authorities section
```

docs/reference/sources_yaml.md
```diff
@@ -0,0 +1,118 @@
...
+doc-topic: [operator-manual, sources.yaml]
+---
+
+<head>
+  <link rel="canonical" href="https://docs.kubewarden.io/reference/sources_yaml"/>
+</head>
+
+## The `sources.yaml` file
+
+You can tune the push-pull behavior of `kwctl` and `policy-server` using the `sources.yaml` file.
```
viccuad, line 18: Not sure, maybe it's worth it to move the paragraph about how to consume the `sources.yaml` out of the reference page and into the howto.

viccuad (approved):
LGTM, little quibbles.

docs/reference/sources_yaml.md
jhkrug: I think it's ok there. For when you land on the ref page and wonder what you can do with the material there. Also, ok on a how-to as well. We can change in the light of experience.

PR #372, opened 2024-03-11
Install cert-manager via Helm in quickstart

viccuad:
## Description

Upstream docs don't wait via `kubectl wait` anymore (see [here](https://cert-manager.io/docs/installation/kubectl/)), and we know that installing via Helm waits correctly. We are also testing the cert-manager installation via helm, too.

## Test

Already tested via E2E tests.


jhkrug (approved):
lgtm

flavio (approved):

PR #375, opened 2024-03-15
A better layout than tables in spec/host-capabilities 

### LINKED ISSUE #374: Investigate the table layouts in Ref>PolSpec>HostCapability pages.
They need too much horizontal scrolling. Probably a different approach is needed.

For example: https://docs.kubewarden.io/next/reference/spec/host-capabilities/kubernetes

jhkrug:
Closes #374.

I had noticed that the host-capabilities pages in policy specification reference used tables for layout. Due to the fixed width content, these tables required a lot of horizontal scrolling and appear to be quite ugly.

Maybe something like this approach is better? As developers, which would you prefer? 

I've used sections, rather than tables, and put the code parts in expandable details. Unfortunately, Ctrl-f doesn't work on strings inside unexpanded details sections, so that's not great.

When we have decided on an approach (or decide to keep the current layout), I'll do the rest of the files in this section that have similar layout.

So, experimental, draft PR. Preferences?

New page: https://deploy-preview-375--silly-bunny-8cedd0.netlify.app/next/reference/spec/host-capabilities/signature-verifier-policies

Old page: https://deploy-preview-375--silly-bunny-8cedd0.netlify.app/writing-policies/spec/host-capabilities/signature-verifier-policies

### DIFF
--- a/docs/reference/spec/host-capabilities/02-signature-verifier-policies.md
+++ b/docs/reference/spec/host-capabilities/02-signature-verifier-policies.md
@@ -100,10 +100,7 @@ these are the functions a policy, that verifies signatures, can use:
 
 ### waPC function - `v2/verify`
 
-#### type - `SigstorePubKeyVerify`
-
-<details>
-<summary>Input payload `SigstorePubKeyVerify`</summary>
+#### `SigstorePubKeyVerify` input
 
 ```hcl
 {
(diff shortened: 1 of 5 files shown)


viccuad (approved):
I like this approach, it's clearer and concise, plus no missed info if one forgets to scroll. The only drawback is searching with `ctrl+F` for the payload contents, but I think is a good trade-off.
Approving.

fabriziosestito (approved):
looks great

jhkrug:
Ok, I hall do the rest of the section.

jhkrug:
I decided not to use the `<details>` expander thingie, retaining the ability to Ctrl-F search.

jhkrug:
Moved back to draft while I make some consistency changes.

jvanz (approved):
LGTM

PR #389, opened 2024-04-08
docs: update generated PolicyServer CRD docs with limits and requests

fabriziosestito:
## Description

update generated PolicyServer CRD docs with limits and requests


viccuad (approved):

flavio (approved):
LGTM.

@viccuad are we going to cover these new attributes on the "running kubewarden in production" page?

Issue #394, opened 2024-04-15
Feature: Establish glossary and setup Kubewarden terms variable expansion


viccuad:
From https://github.com/kubewarden/docs/pull/393#discussion_r1562516046.

## Acceptance criteria

- [x] Establish a glossary as a part of the main documentation set. Also done in #396.
- [x] #396 Update docs to substitute terms exclusive to Kubewarden (such as PolicyServer, policy-server, kwctl). They should link to  their reference pages. For example, the CRD docs, a reference glossary page, etc. Example: `[[< policy-server-component >]]`
- Follow https://kubernetes.io/docs/contribute/style/style-guide for those terms that are Kubernetes resources.

jhkrug:
#396 should be what we are looking for.

There is some Docusaurus docs at https://docusaurus.io/docs/api/docusaurus-config#preprocessor

PR #396, opened 2024-04-17
Infrastructure for implementing a Kubewarden controlled vocabulary.

jhkrug:
docusaurus.config.js defines a preprocessor which replaces all instances of the variables defined in `./variables.json` with their values. See the changes in quick-start.md for an example. `[[< policy-server-component >]]' to 'PolicyServer'.

Part of #394.

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -13,10 +13,92 @@ doc-topic: [glossary]
   <link rel="canonical" href="https://docs.kubewarden.io/glossary"/>
 </head>
 
-## PolicyServer {#policy-server-component}
+## A
 
-A 'PolicyServer' is something, something.
+### AdmissionPolicy
 
-## ClusterAdmissionPolicy {#cluster-admission-policy-component}
+A namespace-wide resource. The policy processes only requests targeting the
+namespace where the AdmissionPolicy is defined.
 
-Is something else.
+### Annotation
+
+### Artifact
+
+### Artifacthub
+
+## C
+
+### ClusterAdmissionPolicy
+
+A ClusterAdmissionPolicy defines how policies evaluate requests.
+
+### ClusterPolicyReport
+
+### Context aware
+
+### ContextAwareResource
+
+## I
+
+### Ingress
+
+## K
+
+### Kubewarden
+
+### kwctl
+
+## M
+
+### Mutating
+
+### MutatingWebhook
+
+### MutatingWebhookConfiguration
+
+## O
+
+### OpenPolicyAgent
+
+## P
+
+### Policy
+
+### PolicyReport
+
+### PolicyServer {#policy-server}
+
+A PolicyServer validates incoming requests by executing Kubewarden policies against requests.
+
+### PublicResource
+
+## S
+
+### SecurityContext
+
+### SourceAuthorities
+
+## V
+
+### ValidatingWebhook
+
+### ValidatingWebhookConfiguration
+
+### ValidationRequest
+
+### ValidationResponse
+
+### ValidationWebhook
+
+## W
+
+### waPC
+
+### WASI
+
+### WASM or Wasm?
+
+### Wasmtime
+
+### WebAssembly
+
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -24,23 +24,28 @@ the namespace in which the AdmissionPolicy is defined.
 
 ### ClusterAdmissionPolicy
 
-A ClusterAdmissionPolicy defines how policies evaluate requests.
+An [AdmissionPolicy](#admissionpolicy) which targets cluster-wide resources.
 
 ### ClusterPolicyReport
 
 A [PolicyReport](#policyreport) and a ClusterPolicyReport store results of
-policy scans. Which one is used depends on the scope of the resource.
+policy scans. Which one is used, depends on the scope of the resource.
 
 ## K
 
 ### kwctl
 
-A CLI tool allowing administrators to test policies before applying them to a
-cluster.
+A CLI tool to generate and test Kubernetes YAML files for policy deployment.
 
 ## M
 
 ### MutatingWebhookConfiguration
+
+A
+[Kubernetes resource](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#what-are-admission-webhooks)
+created by the Kubewarden controller to let Kubernetes know where to send an `AdmissionReview`.
+In other words,
+this is how a Kubewarden controller informs Kubernetes where to find a resource mutating policy.
 
 ## P
 
@@ -56,6 +61,11 @@ A PolicyServer validates incoming requests by executing Kubewarden policies agai
 ## V
 
 ### ValidatingWebhookConfiguration
+
+A
+[Kubernetes resource](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#what-are-admission-webhooks)
+created by the Kubewarden controller to let Kubernetes know where to send a `AdmissionReview`.
+In other words, this is how Kubewarden informs Kubernetes where to find a resource validating policy.
 
 ## W
 
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -1,0 +1,22 @@
+---
+sidebar_label: Glossary
+sidebar_position: 90
+title: Glossary
+description: Kubewarden glossary
+keywords: [kubewarden, glossary]
+doc-persona: [kubewarden-all]
+doc-type: [explanation]
+doc-topic: [glossary]
+---
+
+<head>
+  <link rel="canonical" href="https://docs.kubewarden.io/glossary"/>
+</head>
+
+## PolicyServer {#policy-server-component}
+
+A 'PolicyServer' is something, something.
+
+## ClusterAdmissionPolicy {#cluster-admission-policy-component}
+
+Is something else.
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -17,14 +17,8 @@ doc-topic: [glossary]
 
 ### AdmissionPolicy
 
-A namespace-wide resource. The policy processes only requests targeting the
-namespace where the AdmissionPolicy is defined.
-
-### Annotation
-
-### Artifact
-
-### Artifacthub
+A namespace-wide resource. The policy processes only those requests targeting
+the namespace in which the AdmissionPolicy is defined.
 
 ## C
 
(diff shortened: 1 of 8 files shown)


fabriziosestito (approved):

docs/quick-start.md
```diff
@@ -18,11 +18,11 @@ The Kubewarden stack comprises:
 
 * Some `ClusterAdmissionPolicy` resources: this is how policies are defined for Kubernetes clusters
 
-* Some `PolicyServer` resources: representing a deployment of a Kubewarden `PolicyServer`. Your administrator's policies are loaded and evaluated by the Kubewarden `PolicyServer`
+* Some `[[< policy-server-component >]]` resources: representing a deployment of a Kubewarden `[[< policy-server-component >]]`. Your administrator's policies are loaded and evaluated by the Kubewarden `[[< policy-server-component >]]`
```
viccuad, line 21:
```suggestion
* Some `[[< policy-server-component >]]` resources: representing a deployment of a Kubewarden policy-server. Your administrator's policies are loaded and evaluated by the Kubewarden `[[< policy-server-component >]]`
```

Maybe?

As in, the Custom Resource PolicyServer represents deployment of the policy-server image/binary.

docs/quick-start.md
```diff
@@ -97,17 +97,17 @@ helm install --wait -n kubewarden kubewarden-defaults kubewarden/kubewarden-defa
 ```
 
 :::caution
-Since [`v0.4.0`](https://github.com/kubewarden/kubewarden-controller/releases/tag/v0.4.0), a `PolicyServer` resource named `default` will not be created using the `kubewarden-controller` chart.
+Since [`v0.4.0`](https://github.com/kubewarden/kubewarden-controller/releases/tag/v0.4.0), a `[[< policy-server-component >]]` resource named `default` will not be created using the `kubewarden-controller` chart.
```
viccuad, line 100: @kubewarden/kubewarden-developers  can we drop this caution block already?

docs/quick-start.md
```diff
@@ -116,21 +116,21 @@ The default configuration values are sufficient for most deployments. All option
...
-A `PolicyServer` validates incoming requests by executing Kubewarden policies against them.
+A `[[< policy-server-component >]]` validates incoming requests by executing Kubewarden policies against them.
 
-This is the default `PolicyServer` configuration:
+This is the default `[[< policy-server-component >]]` configuration:
 
 ```yaml
 apiVersion: policies.kubewarden.io/v1
-kind: PolicyServer
+kind: [[< policy-server-component >]]
```
viccuad, line 133:
```suggestion
kind: PolicyServer
```

We shouldn't do it inside code blocks, as they are copy-pasted when trying examples.

variables.json
```diff
@@ -0,0 +1,4 @@
+{
+  "variable_name": "variable_value",
+  "policy-server-component": "PolicyServer"
```
viccuad, line 3: Quibble: Maybe better to name the variable `policy-server-customresource`, or `policy-server-cr` (written PolicyServer), in preparation for having a `policy-server-image`, or `policy-server-pod`, or `policy-server-binary` (written policy-server).

viccuad (commented):
Added some review comments.

Could the policy-server-component variable for "PolicyServer" link to a glossary page with a paragraph explanation?
I would like to picture how the different terms can be listed.

jhkrug:
> Added some review comments.
> 
> Could the policy-server-component variable for "PolicyServer" link to a glossary page with a paragraph explanation? I would like to picture how the different terms can be listed.

Yes, that could be done.

variables.json
jhkrug: It's just an example to show how it works, constructing the list of variable_names and variable_values will be quite an intensive joint effort. I think I'll move this to draft, as a reminder to remove the changes to quick-start.md before merging. They come after, the name and values are decided.

docs/quick-start.md
jhkrug: I guess this is where we will, in due course, have [[< policy-server-deployment >]] mapping to policy-server.

docs/quick-start.md
jhkrug: Just using the quick start for examples of how this might work for now.

jvanz (approved):

variables.json
viccuad: fair enough, no need to block on this for this pr. Resolving.

docs/quick-start.md
viccuad: Oh, yep, they would not copy-paste from the markdown source.
I was on the impression that the substitution would be for a "PolicyServer" that is a link to a reference page or something. If that's the case, then the substition shouldn't be part of the yaml block I suppose. If not, disregard.

viccuad:
> Also, we may want to think about TermWeb (I think you can get to it now, via Okta B2C), SUSEs new terminology database, as opposed to a in-house glossary.

I would prefer to link to a local glossary part of our docs. We are part of CNCF yaddayadda, plus ideally we should ship the docs somehow, in an airgap installation, without needing to hit external URLs.

docs/quick-start.md
jhkrug: Yeah, might be getting a little more complex. Would need variables policy-server-component and policy-server-component-in-a-code-block, or the variable replacement code needs to not add a link to a glossary if inside a code block. Hmmm.

jhkrug:
A discussion. Considering, for the moment, just the term PolicyServer, capitalized in the Kubernetes style, it is a Kubewarden 'component'.

So, how should we be seeing PolicyServer in the documentation? It appears in both inline code-blocks and fenced code-blocks. When it appears in the text, in a sentence, is it always in an inline code-block? If not, should it be enclosed in either single or double quotation marks? Are we trying to indicate a difference between a PolicyServer in the text, and a PolicyServer in a code-block? I hope not, and I don't think we are. 

So, if we want PolicyServer in the text to be both in an inline code-block for formatting/semantic reasons and we also want it to be a tooltip and a link to the glossary then the expansion of `[[> policy-server-component <]]` can be configured to look exactly like it currently does in an inline code-block, but colored a little differently to indicate it's a tooltip/link.

This is moving away from the semantic meaning of the markdown text which I'm reluctant to do.

Hmm, more thinking needed.

jhkrug:
Take a look at https://deploy-preview-396--silly-bunny-8cedd0.netlify.app/next/quick-start the first paragraph of bullet points for a better idea of how it might look.

Also please take a look at the glossary. Some terms are probably unnecessary there and some need to be added. If you want to contribute some pithy text for any of the entries, cool!

jvanz:
I think the glossary is great! We are moving in the right direction. I'm not sure where we use all the terms, but I would say we can reduce the initial terms count there just for the ones more directed related to Kubewarden.  Keeping only:

- AdmissionPolicy
- ClusterAdmissionPolicy
- ClusterPolicyReport
- kwctl
- MutatingWebhookConfiguration
- PolicyReport
- PolicyServer
- ValidatingWebhookConfiguration
- waPC
- WASI
- Wasm
- Wasmtime


Or maybe keeping the term used in the Quick Start for now. I don't know... I'm just wondering about that.

docs/glossary.md
```diff
@@ -0,0 +1,77 @@
...
+
+A [PolicyReport](#policyreport) and a ClusterPolicyReport store results of
+policy scans. Which one is used depends on the scope of the resource.
+
+## K
+
+### kwctl
+
+A CLI tool allowing administrators to test policies before applying them to a
+cluster.
```
jvanz, line 39: The tools also allows the user to generate the Kubernetes yaml files used to deploy the policies.

docs/glossary.md
```diff
@@ -0,0 +1,77 @@
...
+### AdmissionPolicy
+
+A namespace-wide resource. The policy processes only those requests targeting
+the namespace in which the AdmissionPolicy is defined.
+
+## C
+
+### ClusterAdmissionPolicy
+
+A ClusterAdmissionPolicy defines how policies evaluate requests.
```
jvanz, line 27: This is the same of the `AdmissionPolicy`. But this targets cluster-wide resources.

docs/glossary.md
```diff
@@ -0,0 +1,77 @@
...
+## K
+
+### kwctl
+
+A CLI tool allowing administrators to test policies before applying them to a
+cluster.
+
+## M
+
+### MutatingWebhookConfiguration
```
jvanz, line 43: This is the [Kubernetes resource](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#what-are-admission-webhooks) created by the Kubewarden controller to let Kubernetes knows where it should sent the `AdmissionReview`. In other words, this is how Kubewarden controller teach Kubernetes where to find the policy which could mutate resources before deploy them in the cluster

docs/glossary.md
```diff
@@ -0,0 +1,77 @@
...
+A PolicyReport and a [ClusterPolicyReport](#clusterpolicyreport) store results of
+policy scans. Which one is used depends on the scope of the resource.
+
+### PolicyServer {#policy-server}
+
+A PolicyServer validates incoming requests by executing Kubewarden policies against requests.
+
+## V
+
+### ValidatingWebhookConfiguration
```
jvanz, line 58: This is the [Kubernetes resource](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#what-are-admission-webhooks) created by the Kubewarden controller to let Kubernetes knows where it should sent the `AdmissionReview`. In other words, this is how Kubewarden controller teach Kubernetes where to find the policy which  validates resources before deploy them in the cluster

jvanz (changes requested):

docs/glossary.md
jhkrug: Thanks.

jhkrug:
Ready for review again. Keeping it in draft, I'm still not totally convinced. See what you think. The quick start page is the only one modified (so far) to use this idea of the controlled vocabulary.

jvanz (commented):
LGTM

The glossary items and the position of the tooltips looks fine to me. I'm just not sure about the colors. But I do not know how to improve it.

jhkrug:
> LGTM
> 
> The glossary items and the position of the tooltips looks fine to me. I'm just not sure about the colors. But I do not know how to improve it.

Yes, it makes the page quite busy. I'm tempted to try no color in the text, so no visual indication at all that there is a tooltip. So, once you know there may be one, you can use it if you want to. It's not quite as invisible as it seems. It's surprising how much we use a mouse/pointer when reading.

jhkrug:
Ok, I think this is the final version. Much more subtle.

jvanz:
> Ok, I think this is the final version. Much more subtle.

That's it! I like it

jvanz (approved):

jhkrug:
I think this can be merged. It could be improved further, by adding some code to reduce duplication in variables.json. For a future iteration, when docs infrastructure is decided on.

PR #402, opened 2024-05-15
Use the CNCF gtag for analytics

jhkrug:
Note to self: Check that it's working after merge.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -124,7 +124,6 @@ module.exports = {
           anonymizeIP: true,
         },
         gtag: {
-          // trackingID: "395802740", // Google Analytics tracking ID for SUSE Rancher
           trackingID: "G-PSW07XK7TM", // Google Analytics tracking ID for CNCF
           anonymizeIP: true,
         },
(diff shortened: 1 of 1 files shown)


docusaurus.config.js
```diff
@@ -124,7 +124,8 @@ module.exports = {
           anonymizeIP: true,
         },
         gtag: {
-          trackingID: "395802740",
+          // trackingID: "395802740", // Google Analytics tracking ID for SUSE Rancher
```
viccuad, line 127: I think it's fine to drop the old one as comment, we have in history.

docusaurus.config.js
```diff
@@ -124,7 +124,7 @@ module.exports = {
           anonymizeIP: true,
```
chalin, line 124: Google Analytics 4 (GA4) replaced Google Universal Analytics (UA) effective 1 July 2023. See https://support.google.com/analytics/answer/11583528, so you can drop the UA ID on line 123 above (well, remove the entry from 122-125). /cc @jvanz

PR #411, opened 2024-06-13
fix: enable right policy-reporter UI ingress.

jvanz:
## Description

In the audit scanner the ingress configuration provided enabled the ingress to the policy-reporter service not for its UI. As stated in the documentation. This commit fix that by updating the example ingress configuration enabling the UI one.

Fix https://github.com/kubewarden/helm-charts/issues/463


flavio:
Tested, it works

PR #433, opened 2024-08-01
Refresh OPA Gatekeeper comparison with CEL

viccuad:
## Description

Refresh OPA Gatekeeper comparison with CEL.


jhkrug (changes requested):
a couple of minor suggestions

viccuad:
Thanks, applied!

jhkrug (approved):
lgtm

PR #441, opened 2024-08-16
Fix the name of the get_resource operation

brownz11:
## Description

Fix Doc issue, the operation is called "get_resource" singular per https://github.com/kubewarden/policy-evaluator/blob/3cd66b932b199037e677e3e204d4d9742e23edc8/src/runtimes/callback.rs#L276

## Test

None

### Tradeoff

None

### Potential improvement

None


viccuad (approved):
Many thanks for the fix!

PR #444, opened 2024-08-27
fix: We don't expect kwctl to output a json anymore

viccuad:
## Description

Relates to
https://github.com/kubewarden/kubewarden-controller/issues/838.


jhkrug (approved):
lgtm

PR #453, opened 2024-09-13
feat(docs): explanation page about policy groups.

jvanz:
## Description

Adds a new page in the documentation explaining how the policy group types works.

Fix #448

PR #454, opened 2024-09-13
feat(docs): howto use policy groups.

### LINKED ISSUE #449: PolicyGroup: add howto page
Write a new page under the "Howtos" section that explains how to make use of the PolicyGroup feature.

We could copy examples from the examples made inside of the [RFC](https://github.com/kubewarden/rfc/blob/main/rfc/0021-policy-group.md)

jvanz:
## Description

Adds a new Howto page showing how to use the policy groups.

Fix #449

### DIFF
--- a/docs/howtos/policy-groups.md
+++ b/docs/howtos/policy-groups.md
@@ -1,0 +1,123 @@
+---
+sidebar_label: Policy Groups
+sidebar_position: 36
+title: How to use policy groups
+description: How to use Kubewarden policy groups
+keywords: [kubewarden, policy groups, clusteradmissionpolicygroup, admissionpolicygroup]
+doc-persona: [kubewarden-operator]
+doc-type: [howto]
+doc-topic: [explanations, policy-group]
+---
+
+<head>
+  <link rel="canonical" href="https://docs.kubewarden.io/howtos/policy-groups"/>
+</head>
+
+:::tip
+Before working with Policy Groups, consult the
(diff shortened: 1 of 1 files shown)


jhkrug (changes requested):
Some suggestions for your consideration.

flavio:
I would even close this PR. It's basically a shorted version of the explanation page provided by https://github.com/kubewarden/docs/pull/453

@jhkrug: what do you think, am I missing some nuance?

jhkrug:
I thought they were quite similar, other than being in explanations and howtos. Maybe that's the distinction. Maybe another iteration of both, focusing on a better separation of the content into howto and explanation might be worth doing.

jhkrug:
I think the whole first part could be replaced with:

Prior to working with Policy Groups, consult the [explanation](link) for them.

Then another look at what remains. The example is very similar to that in explanations. Say, something, like:

Using the example from [the explanation of Policy Groups](link to the example) use these commands to implement (or define) it.

Commands

Output.

That should do. The Context aware rules section is already covered in Context-Aware policies in the explanation (I think) so could be dispensed with.

What do you think?

flavio:
@jhkrug I've made the changes you described

jhkrug (changes requested):
Two little suggestions to consider, but otherwise lgtm.

flavio:
@jhkrug thanks for the last review

PR #456, opened 2024-09-26
fix(docs): removes policy group warnings from docs.

jvanz:
## Description

The policy group does not emit warnings anymore. Therefore, this commit updates the documentation about policy groups removing the mentions to the warning emitted by the evaluation.

Fix https://github.com/kubewarden/policy-server/issues/927


jhkrug (approved):
lgtm

PR #471, opened 2024-12-02
Remove no longer existing OCI registry link to bundle.bar

jhkrug:


viccuad (approved):
LGTM. For these cases I'm also ok removing it only on `next/`.

flavio (approved):
thanks for the fix

jhkrug:
> LGTM. For these cases I'm also ok removing it only on `next/`.

Yes, it felt like re-writing history. I couldn't work out when bundle.bar ceased to be which may have provided more guidance. Anyway ... it improves the accessibility/quality score that's getting generated for the product docs.

PR #472, opened 2024-12-05
Add reference/artifacts page

viccuad:


kkaempf (commented):
Hmm, this PR seems to mix multiple changes beyond adding the artifacts page ?! 🤔

flavio (changes requested):
The new content looks good, but it looks like you built the PR on top of another one. Can you rebase against the main branch?

flavio (approved):
thanks!

PR #489, opened 2025-01-15
feat(ci): CI to update CLI reference docs.

jvanz:
## Description

Adds a new CI job to automatically updates the CLI reference docs.

Fixes https://github.com/kubewarden/kwctl/issues/851


updatecli/update-cli-ref-docs.yaml
```diff
@@ -0,0 +1,6 @@
+policies:
+  - name: Update kwctl CLI reference documentation
+    policy: ghcr.io/jvanz/updatecli/policies/file:0.2.0
```
jvanz, line 3: This must be updated when the policy got accepted in the updatecli org

https://github.com/updatecli/policies/pull/41

fabriziosestito (approved):

updatecli/values.d/update-cli-ref-docs.yaml
```diff
@@ -0,0 +1,26 @@
+---
+# Values.yaml contains settings that be used from Updatecli manifest.
+sourceFiles: 
+  - "cli-docs.md"
+targetFiles: 
+  - "docs/reference/kwctl-cli.md"
```
viccuad, line 6: Is the copied file `kwctl-cli.md` consumable by the docs directly? They may need a header, and this header may change in the future if we move away from docusaurus.

viccuad (changes requested):
Is the copied file `kwctl-cli.md` consumable by the docs directly? They may need a header, and this header may change in the future if we move away from docusaurus.

I think we should add a header file that we append the `kwctl-cli.md`. I suppose this could be an updatecli task.

jvanz:
> Is the copied file `kwctl-cli.md` consumable by the docs directly? They may need a header, and this header may change in the future if we move away from docusaurus.
> 
> I think we should add a header file that we append the `kwctl-cli.md`. I suppose this could be an updatecli task.

Yes, it works with no header. But you're right. I've forgot to add the header. Let me fix it.

jvanz:
@viccuad this is an example of a [PR](https://github.com/jvanz/docs/pull/26) created by this changes

updatecli/updatecli.d/headers.yaml
```diff
@@ -0,0 +1,62 @@
...
+    spec:
+      owner: {{ requiredEnv .scms.owner }}
+      repository: {{ .scms.repository }}
+      branch: {{ .scms.branch }}
+      user: {{ requiredEnv .scms.owner }}
+      username: {{ .scms.username }}
+      token: {{ requiredEnv .scms.token }}
+
+sources:
+# {{ range $index, $file := .files }}
```
flavio, line 18: I'm confused about this line and other ones, are these comments or did you mean to have something like that?

```
#{{ range $index, $file := .files }}
```

flavio:
BTW, I like the approach taken!

updatecli/updatecli.d/headers.yaml
olblak: I like this approach because using the comment allow us to use go template, while keeping the advantage of having a compliant yaml file.

Differently said:
```
{{ range $index, $file := .files }}
```
Would also work, but your IDE would complain that it's not a valid yaml file and the Updatecli jsonschema wouldn't be loaded loosing Updatecli doc and auto-completion from IDE supporting jsonschema store

viccuad (approved):
LGTM, thanks for tackling the header!

flavio (approved):

PR #544, opened 2025-02-19
Update policy-reporter content to v3.0

viccuad:
## Description

Part of https://github.com/kubewarden/kubewarden-controller/issues/991


jhkrug (changes requested):
A few minor suggestions for your consideration. Thanks.

I think maybe we should have some policy for formatters at some point. Sometimes it's hard to pick out the content changes from formatting by editors/tool setup. I think we all do this sometimes. I'll come up with something when I have a chance.

viccuad:
> I think maybe we should have some policy for formatters at some point. Sometimes it's hard to pick out the content changes from formatting by editors/tool setup. I think we all do this sometimes. I'll come up with something when I have a chance.

Totally fair, sorry.
Currently I have https://prettier.io formatter installed on my editor, as we use it for other repos too. I put the formatting changes in a separate commit, so if in doubt, it's good to review per commit changes instead of files.

jhkrug:
> > I think maybe we should have some policy for formatters at some point. Sometimes it's hard to pick out the content changes from formatting by editors/tool setup. I think we all do this sometimes. I'll come up with something when I have a chance.
> 
> Totally fair, sorry. Currently I have https://prettier.io formatter installed on my editor, as we use it for other repos too. I put the formatting changes in a separate commit, so if in doubt, it's good to review per commit changes instead of files.

Yeah, I kinda like prettier mostly but I hate what it does to md tables. It's fine when you have small text and wide screens. If you have giant comedy size text and like more or less 80 col widths like me then it's a pain. I know it's possible to have ignore directives but don't really want them scattered around either. Maybe I just have to bite the bullet here. :-)

PR #539, opened 2025-01-31
CNCF recommendations - Introductory pages review for active voice, language and some formatting.

jhkrug:
As per CNCF recommendations.

PR #549, opened 2025-03-03
fix(quick-start): correctly render variables

agilgur5:
## Description

In the quick-start (`/quick-start`), a new line was added in between the [required syntax](https://github.com/kubewarden/docs/blob/a1ac1a8b84727bbe7c1f9a2b6956dc3a83688f98/js-lib/docusaurus-variables.js#L11) for two variables, making them erroneously render as plain text
  - move the whole variable to the next line to properly render it, matching how it was before https://github.com/kubewarden/docs/pull/539#discussion_r1976789946
  
<details><summary>Here's how it looked like <i>before</i> this change:</summary>

Screenshot 2025-03-02 at 8 50 14 PM

</details>

## Test

```sh
yarn start &
open http://localhost:3000/quick-start
```

### Tradeoff

n/a

### Potential improvement

n/a

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -33,9 +33,9 @@ your administrator's policies.
 
 - Some [[< admission-policy >]] resources: policies for a defined namespace.
 
-- A deployment of a `kubewarden-controller`: this controller monitors the [[<
-cluster-admission-policy >]] resources and interacts with the Kubewarden [[<
-policy-server >]] components.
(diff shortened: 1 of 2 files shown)


viccuad (approved):
Woops, good catch. Many thanks!

flavio:
thanks for the fix!

jhkrug:
Thank you.

PR #583, opened 2025-03-25
fix: mtls howto

flavio:
Incorporate feedback from @pjbgf:

- Increase the expiration time of the root CA and the client certificate
  issued by that
- Provide a simple openSSL configuration when creating the root CA and
  the client certificate
- Increase the RSA key size
- Fix a typo about the location of the `admission.yaml` file

Thanks to @pjbgf for his review!


viccuad (approved):
LGTM

PR #591, opened 2025-04-22
This may fix the build problems for Netlify (maybe).

jhkrug:
Netlify builds are failing. Increasing resources for node may help.


jvanz (approved):
The CI is green and the preview seems to work. Therefore, I think that everything is fine! :)

viccuad (approved):

PR #605, opened 2025-05-13
docs: expand security hardening with PSA

flavio:
Clarify the section about PSA, previously it was kinda conflated into the pod security context section, which was wrong.

[Link to the new section](https://deploy-preview-605--docs-kubewarden-io.netlify.app/next/howtos/security-hardening#pod-security-admission)


jhkrug (approved):
lgtm

Issue #477, opened 2024-12-17
Documentation improvements as per CNCF recommendations


jhkrug:
CNCF recommendations report is [here](https://docs.google.com/document/d/1shAoi7TJOtuW8z6heR_sUhv3uE7Gk6k6argd_zffaZk/edit?tab=t.0): 

This issue lays out the needed/suggested tasks:

## Specific

- [ ] Documentation layout description page
- [ ] Provide a page with a description of personas using and developing Kubewarden, at whom the documentation is aimed.
  - Provide a map/list of pages relevant to each persona. Continues/extends work began in #400 and #425.
  - Each page should have a statement regarding the personas the page is aimed at?
  - For each persona provide: their likely experience, Kubernetes knowledge, goals in using the docs, expected learnings.
- [ ] Kubewarden what is/introduction page review
  - New user section. If you are new to KW ... blah, blah. 
  - Core features and benefits. What problem(s) does KW solve? The value proposition?
  - Link to and review/improve the explanation/architecture (referred to as the 'Overview' in the CNCF critique) page
  - Provide a recent developments and future roadmap section. Possibly point to an additional page.
- [ ] Quick start page review
  - Link to the improved architecture document to provide context
  - Where can KW be installed? On prem? Cloud? Well, it's really wherever your K8s clusters are
  - Helm charts and their role in installation. More context and links to further information
  - Better description of prerequisites and their configuration
  - Common issues and errors
  - Validate copy/paste installation of basic installation on basic K8S. How deep to go?
  - Review main components section
  - Format and legal value for attributes table.
  - Further worked examples for differing scenarios?
- [ ] New main docs section 'Policy management (& development)'?
  - Management
  - Development tutorials (Rust, etc.)

## General

- [ ] Review for active voice
- [ ] Style guide consistency, currently Microsoft(y), but SUSE docs have their own style guide. Which is quite different.

PR #609, opened 2025-05-19
Updates to the upgrade path documentation.

jhkrug:
Part of the work on #477.

Remove the upgrade path diagram and commentary. It's about 20 versions old.

Does what remains suffice? I'm assuming we have had no complex upgrade paths recently?

Language improvements as per #477.

### DIFF
--- a/docs/reference/upgrade-path.md
+++ b/docs/reference/upgrade-path.md
@@ -17,7 +17,7 @@ doc-topic: [operator-manual, upgrade-path]
 
 The Kubewarden project uses [Semantic versioning](https://semver.org/)
 to define the "stack" version of all its components: the version follows the
-`MAJOR.MINOR.PATCH` pattern. The supported version is only the latest release.
+`MAJOR.MINOR.PATCH` pattern. The supported version is the latest release.
 
 The Kubewarden components that follow the rules for the "stack" version are:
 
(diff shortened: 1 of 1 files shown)


docs/reference/upgrade-path.md
```diff
@@ -89,20 +62,20 @@ custom ones) and `kwctl`.
...
 ## Downgrades
 
-Downgrades are **not** supported, and haven't been tested. Nevertheless,
-there's a reasonable expectation that they should work.
+Downgrades are **not** supported, and Kubewarden does not test them.
+Nevertheless, there's a reasonable expectation that they should work.
 
 ## SDKs, policies
 
 Policy SDKs for the different languages, and policies maintained by the
```
dariavladykina, line 70:
```suggestion
Policy SDKs for the different languages and policies maintained by the
```

dariavladykina (commented):
A minor comma spotted. Thanks!

docs/reference/upgrade-path.md
```diff
@@ -39,46 +39,19 @@ Therefore, if the `kubewarden-controller` version running is `1.1.x`, the
...
-  1.1.2-->1.2.1(1.2.1);
-  1.2.0-->1.2.1;
-  linkStyle 6 color:brown,stroke-width:2px,stroke-dasharray: 3 5
-  linkStyle 7 color:red,stroke-width:2px,stroke-dasharray: 3 5
-```
-
-<figcaption>Upgrade path support graph</figcaption>
-</figure>
-
-For example, the user is allowed to upgrade components from version `1.1.10` to
```
viccuad, line 75: I have to say, that I liked the examples and the diagram to help understand it. I have no problem with old versions featured there, the important parts are the bumps of MAJOR, MINOR, and PATCH . If anything, it signifies that we have stuck to the same advertised behavior since 1.0.0.

jhkrug: > I have to say, that I liked the examples and the diagram to help understand it. I have no problem with old versions featured there, the important parts are the bumps of MAJOR, MINOR, and PATCH . If anything, it signifies that we have stuck to the same advertised behavior since 1.0.0.

Ahh, ok. I always just assumed that these were 'real' KW versions and there was a mildly complex scenario involving them. So, it is actually an example of how to do upgrades from a variety of versions. 

OK, no problem. I'll put that back in. I want to change so that the diagram comes after the commentary and that the diagram title includes the work 'example'. Does that seem ok @viccuad ?

jhkrug:
OK, how does that seem now?

jvanz (approved):

docs/reference/upgrade-path.md
viccuad: That seems indeed better. Thanks!

PR #612, opened 2025-05-21
Fix link to Jaeger Kubernetes operator.

jhkrug:
Payback backport from product docs link checking performed by docs team with SiteImprove.


flavio (approved):
Thanks for the fix

PR #614, opened 2025-05-21
struct.Pod ref broken link fix.

jhkrug:
The original link is to v0.11.x which is now 404. I've updated it to v0.25.x. Is that OK? Lots of versions in between. I don't know if it makes difference or not.


jvanz:
> The original link is to v0.11.x which is now 404. I've updated it to v0.25.x. Is that OK? Lots of versions in between. I don't know if it makes difference or not.

Yes, it's fine. We recently updated our dependency to use the v0.25 version.

PR #633, opened 2025-07-07
feat(can-i): kubernetes/can-i capability documentation.

### LINKED ISSUE #629: Document new host capability to check service account permissions
Update the documentation adding information about the new host capability that allow users checking service account permissions

### Acceptance criteira
- Update reference documentation adding a the new `kubernetes/can-i` host capability

jvanz:
## Description

Adds the reference documentation for the `kubernetes/can-i` host capability.

Fix #629

### DIFF
--- a/docs/reference/spec/host-capabilities/06-kubernetes.md
+++ b/docs/reference/spec/host-capabilities/06-kubernetes.md
@@ -224,7 +224,7 @@ Example of an operation blocked by some authorization plugin:
   "allowed": false, // Allowed is required. True if the action would be allowed, false otherwise.
   "denied": true,   // Denied is optional. True if the action would be denied, otherwise false. If both allowed is false and denied is false, then the authorizer has no opinion on whether to authorize the action. Denied may not be true if Allowed is true.
   "evaluationError": "", // EvaluationError is an indication that some error occurred during the authorization check. It is entirely possible to get an error and be able to continue determine authorization status in spite of it. For instance, RBAC can be missing a role, but enough roles are still present and bound to reason about the request.
-  "reason": "" // Reason is optional. It indicates why a request was allowed or denied.
+  "reason": "User \"john\" cannot create resource \"pods\" in API group \"\" in the namespace \"development\": no RBAC rule matched" // Reason is optional. It indicates why a request was allowed or denied.
 }
 ```
 


docs/reference/spec/host-capabilities/06-kubernetes.md
```diff
@@ -14,35 +14,46 @@ doc-topic:
...
+- `list_resources_by_namespace` : Given a resource type and a namespace, list
+  all the resources of that type that are defined in it. This cannot be used to
+  list cluster-wide resources, like `Namespace`.
+- `list_resources_all`: Given a resource type, list all the resources of that
+  type that are defined inside the whole cluster. This can be used to list
+  cluster-wide resources, like `Namespace`.
+- `get_resource`: Find the exact resource identified by the given resource
+  type, given name and an optional namespace identifier.
+- `can_i`: Allow policy authors to send `SubjectAccessReview` object to Kubernetes
+  authorization API to verify user permissions
```
viccuad, line 36:
```suggestion
  authorization API to verify user permissions.
```

docs/reference/spec/host-capabilities/06-kubernetes.md
```diff
@@ -122,3 +135,49 @@ Use the `list_resources_all` when dealing with cluster-wide resources.
 #### Output
 
 Result of `GET /apis/$api_version/namespaces/$namespace/$kind/$name`
+
+### Operation - `can_i`
+
+This operation determines if a specific user has the permission to perform an
```
viccuad, line 141:
```suggestion
This operation determines if a specific user or group has the permission to perform an
```

docs/reference/spec/host-capabilities/06-kubernetes.md
```diff
@@ -122,3 +135,49 @@ Use the `list_resources_all` when dealing with cluster-wide resources.
...
+  # Disable caching of results obtained from Kubernetes API Server
+  "disable_cache": bool
+}
+```
+
+#### Output
+
+The output is a JSON object that contains the status of the
+`SubjectAccessReviewi` request. This object indicates whether the requested
+action is allowed. For a complete reference of all the fields available in the
```
viccuad, line 174:
```suggestion
action is allowed or not. For a complete reference of all the fields available in the
```

docs/reference/spec/host-capabilities/06-kubernetes.md
```diff
@@ -122,3 +135,49 @@ Use the `list_resources_all` when dealing with cluster-wide resources.
...
+output, see the official
+[`SubjectAccessReviewStatus`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#subjectaccessreviewstatus-v1-authorization-k8s-io)
+documentation.
+
+```hcl
+{
+  "allowed": true
+}
+```
```
viccuad, line 183: I would like to add an example of a rejected request, to signify that the return is more complex than just a bool (can't suggest via UI, nested codeblocks). Something like:


Example of a rejected request:
```hcl
{
    "status": {
        "allowed": false,
        "denied": true,
        "reason": "User does not have permission to perform this action."
    }
}
```

docs/reference/spec/host-capabilities/06-kubernetes.md
flavio: I've updated the suggestion made by @viccuad to include also a link to the authorization docs of k8s. This is kinda of a niche API

docs/reference/spec/host-capabilities/06-kubernetes.md
```diff
@@ -122,3 +135,49 @@ Use the `list_resources_all` when dealing with cluster-wide resources.
...
+    "kind": "SubjectAccessReview",
+    "spec": {
+      "user": "system:serviceaccount:mynamespace:myuser",
+      "resourceAttributes": {
+        "namespace": "",
+        "verb": "list",
+        "resource": "secrets"
+      }
+    }
+  } ,
```
flavio, line 164: I would not embed the actual Kubernetes object into our request:

- It ties us to a specific version of a Kubernetes resource
- We send extra useless data between the host and the guest (it's a silly optimization thing, but still)
- We are not consistent with the naming of the variables. This is ugly and error prone. I know this is something the end user is not seeing, but someone writing a SDK could run into issues because some things are camelCase (`apiVersion`) while others are snake case (`subject_access_review`).

I would create our own payload that is snake case and contains only the data we need.

docs/reference/spec/host-capabilities/06-kubernetes.md
flavio: Let's make sure we explain the different meaning, since this is not something immediate to understand

docs/reference/spec/host-capabilities/06-kubernetes.md
jvanz: Opened an PR to remove the wrapper object: https://github.com/kubewarden/policy-evaluator/pull/721

docs/reference/spec/host-capabilities/06-kubernetes.md
```diff
@@ -122,3 +138,83 @@ Use the `list_resources_all` when dealing with cluster-wide resources.
...
+:::
+
+Example of an allowed operation. Some of the authorization plugins allows the operation:
+
+```hcl
+{
+  "allowed": true, // Allowed is required. True if the action would be allowed, false otherwise.
+  "denied": false, // Denied is optional. True if the action would be denied, otherwise false. If both allowed is false and denied is false, then the authorizer has no opinion on whether to authorize the action. Denied may not be true if Allowed is true.
+  "evaluationError": "EvaluationError is an indication that some error occurred during the authorization check. It is entirely possible to get an error and be able to continue determine authorization status in spite of it. For instance, RBAC can be missing a role, but enough roles are still present and bound to reason about the request.",
+  "reason": "Reason is optional. It indicates why a request was allowed or denied."
```
viccuad, line 194: I would prefer all fields marked with a comment for consistency ( `// Reason is optional`).

docs/reference/spec/host-capabilities/06-kubernetes.md
```diff
@@ -122,3 +138,83 @@ Use the `list_resources_all` when dealing with cluster-wide resources.
...
+```
+
+Example of an operation blocked by some authorization plugin:
+
+```hcl
+
+{
+  "allowed": false,
+  "denied": true,
+  "evaluationError": "EvaluationError is an indication that some error occurred during the authorization check. It is entirely possible to get an error and be able to continue determine authorization status in spite of it. For instance, RBAC can be missing a role, but enough roles are still present and bound to reason about the request.",
```
viccuad, line 205: I would like to have some real example strings for `evaluationError` and `reason` (even empty), rather than a description.

docs/reference/spec/host-capabilities/06-kubernetes.md
```diff
@@ -200,10 +221,10 @@ Example of an operation blocked by some authorization plugin:
...
-  "reason": "Reason is optional. It indicates why a request was allowed or denied."
+  "allowed": false, // Allowed is required. True if the action would be allowed, false otherwise.
+  "denied": true,   // Denied is optional. True if the action would be denied, otherwise false. If both allowed is false and denied is false, then the authorizer has no opinion on whether to authorize the action. Denied may not be true if Allowed is true.
+  "evaluationError": "", // EvaluationError is an indication that some error occurred during the authorization check. It is entirely possible to get an error and be able to continue determine authorization status in spite of it. For instance, RBAC can be missing a role, but enough roles are still present and bound to reason about the request.
+  "reason": "" // Reason is optional. It indicates why a request was allowed or denied.
```
viccuad, line 227:
```suggestion
  "reason": "User \"john\" cannot create resource \"pods\" in API group \"\" in the namespace \"development\": no RBAC rule matched" // Reason is optional. It indicates why a request was allowed or denied.
```

flavio (approved):
please merge it once the feedback from @viccuad is incorporated

PR #636, opened 2025-07-10
docs: add caching notes for signature verifier and kubernetes capabilities

### LINKED ISSUE #634: Document which host capabilities have cache
Just as the OCI capabilities document the usage of a cache [here](https://docs.kubewarden.io/reference/spec/host-capabilities/container-registry#caching), document this for others using a cache.

### Acceptance criteria

Look at policy-evaluator to see which capabilities have a cache [here](https://github.com/kubewarden/policy-evaluator/blob/faef8b1a4ab767cbb1e23224e8f5907f56bec3e6/src/callback_handler.rs#L349), and add a mention on their reference docs [here](https://docs.kubewarden.io/reference/spec/host-capabilities/intro-host-capabilities).

esosaoh:
## Description

Fix #634


jhkrug (changes requested):
Minor suggestions for consideration. Thanks.

In general, use  words for numbers >= 10. 

I don't think the emphasis was necessary. It wasn't in the first change, so why in the second? Just looking for consistency.

PR #665, opened 2025-09-02
Explain new Policy `spec.timeoutEvalSeconds` feature

viccuad:
## Description

Fix https://github.com/kubewarden/kubewarden-controller/issues/1190


jvanz (changes requested):
I think the docs looks fine. But I think adding an example of the policy configuration in the `Per policy` section similar to what is done in the policy server configuration example would be great.

docs/reference/policy-evaluation-timeout.md
```diff
@@ -53,20 +54,39 @@ deadlocks.
...
-
 ## Configuration
 
-Policy evaluation timeout is a configuration option of Policy Server enabled by
-default. Interruption of a request evaluation takes place after 2 seconds.
+All values of the per-Policies evaluation timeouts, the per-Policy-Server
+evaluation timeout, and the webhook timeouts are validated so they all are
+within acceptable values between each other. For example, it is not possible to
+set a Policy evaluation timeout value that is higher than the Kubernetes'
+webhook timeout.
```
dariavladykina, line 63:
```suggestion
All values of the per-Policy evaluation timeouts, the per-Policy-Server
evaluation timeout, and the webhook timeouts are validated so they all are
within acceptable values of each other. For example, it is not possible to
set a Policy evaluation timeout value that is higher than the Kubernetes'
webhook timeout.
```

docs/reference/policy-evaluation-timeout.md
```diff
@@ -33,15 +32,17 @@ language [Rego](../tutorials/writing-policies/rego/01-intro-rego.md). Both
...
+When using a traditional,
+[Turing-complete](https://en.wikipedia.org/wiki/Turing_completeness)
+programming language, it's possible to have issues like:
 
-* [infinite loops](https://en.wikipedia.org/wiki/Infinite_loop)
-* [deadlocks](https://en.wikipedia.org/wiki/Deadlock).
-* slow running code lacking optimizations
-* computationally intense operations
+- [Infinite loops](https://en.wikipedia.org/wiki/Infinite_loop)
+- [Deadlocks](https://en.wikipedia.org/wiki/Deadlock).
```
jhkrug, line 40:
```suggestion
- [Deadlocks](https://en.wikipedia.org/wiki/Deadlock)
```

docs/reference/policy-evaluation-timeout.md
```diff
@@ -53,20 +54,74 @@ deadlocks.
...
+All values of the per-Policy evaluation timeouts, the per-Policy-Server
+evaluation timeout, and the webhook timeouts are validated so they all are
+within acceptable values of each other. For example, it is not possible to
+set a Policy evaluation timeout value that is higher than the Kubernetes'
+webhook timeout.
+
+### Per Policy
+
+Starting with Kubewarden v1.29.0, every Kubewarden Policy can set its own
+timeout value via its `spec.timeoutEvalSeconds` attribute. Not to confuse with
```
jhkrug, line 68:
```suggestion
timeout value via its `spec.timeoutEvalSeconds` attribute. This is not to be confused with
```

docs/reference/policy-evaluation-timeout.md
```diff
@@ -103,21 +158,23 @@ Kubewarden is a [webhook](https://en.wikipedia.org/wiki/Webhook) implementation
...
 API server doesn't wait forever. After a certain amount of time, it considers
 the request to have timed out.
 
-Quoting the [Kubernetes documentation](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts):
+Quoting the [Kubernetes
+documentation](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts):
 
 > Because webhooks add to API request latency, they should evaluate as quickly
-> as > possible. `timeoutSeconds` allows configuring how long the API server
+> as possible. `timeoutSeconds` allows configuring how long the API server
```
jhkrug, line 171:
```suggestion
> as possible. Setting `timeoutSeconds` configures how long the API server
```

jhkrug (changes requested):
Tiny suggestions. Thanks.

PR #683, opened 2025-10-06
Update reusing-vap tutorial for paramKind,paramRef

viccuad:
## Description

Part of https://github.com/kubewarden/cel-policy/issues/161


jvanz (changes requested):
LGTM. But there are some minor fixes required. 

When using the params. The object in the `params` variable is the kubernetes resource itself. Therefore, as you are using a `ConfigMap` you must access the `data` field to get the value that you are looking for. ;)

jvanz (approved):

PR #686, opened 2025-10-20
Update docs for the migration from PolicyReports to OpenReports.io

viccuad:
## Description

Fix https://github.com/kubewarden/audit-scanner/issues/596

Left the URL slug for `explanations/audit-scanner/policy-reports` untouched, I think it is better than `reports`, and it matches with `policy-reporter` and so.

## Test

Tested locally, policy-reporter-ui included.


viccuad:
Thanks, good catches! Did another search round and couldn't find anything else.

PR #689, opened 2025-10-21
Use the SUSE font.

jhkrug:
Brand management will like.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -154,4 +154,28 @@ module.exports = {
       },
     ],
   ],
+  headTags: [
+    {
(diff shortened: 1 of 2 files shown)


jhkrug:
> I'm trying to review this but I'm not sure how to do it. Should this change produce some visible change? The only very small difference that I noticed comparing with the current docs website is in the footer. Is that?
> 
> image

The only change is the font. Instead of the standard Docusaurus font, whatever that was named, it now uses the SUSE font for both variable and monospace text. https://brand.suse.com/design-language#typography

Old on the left, new on the right.

image

PR #710, opened 2026-02-02
Update airgap info with new imagelist.txt workflow

viccuad:
## Description

Part of https://github.com/kubewarden/helm-charts/issues/905

This includes instructions for using the `imagelist.txt` and `policylist.txt` included inside of the Helm charts, as the <charts>image-list.txt artifact is now dropped from GH releases with https://github.com/kubewarden/helm-charts/pull/922

It also updates some links to the monorepo.

## Test

Tested the bash oneliners that are introduced, but I didn't test the full airgap process.


jhkrug (approved):
lgtm

PR #712, opened 2026-02-13
docs: Add PolicyServer proxy howto page

viccuad:
## Description

Part of https://github.com/kubewarden/kubewarden-controller/issues/1300


viccuad:
> LGTM. But I think we can say that kwctl is also able to consider this environment variables.

I agree here, but we lack a kwctl how-to page where this would fit. At least, the kwctl cli docs are included and will show on the search.

PR #730, opened 2026-04-17
docs: fix quickstart

flavio:
The policy being referenced could not be downloaded.

Plus, use `default` as policy server for the demo policy, otherwise someone doing copy & paste will find the policy in pending state and will be confused by that.


jhkrug (approved):
ok

jvanz (approved):

PR #758, opened 2026-05-20
Set GA running again.

jhkrug:
GA turned off for go-live. setting it running again.

### DIFF
--- a/kw-community-docs-supp-files/partials/head-meta.hbs
+++ b/kw-community-docs-supp-files/partials/head-meta.hbs
@@ -12,6 +12,8 @@ href="https://fonts.googleapis.com/css2?family=SUSE:wght@100..800&display=swap" 
 <!-- <link href="https://fonts.googleapis.com/css?family=Montserrat|Roboto&display=swap" rel="stylesheet"> -->
 <link rel="stylesheet" href="{{uiRootPath}}/css/vendor/light.css">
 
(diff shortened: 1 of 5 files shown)


kw-community-docs-supp-files/partials/head-meta.hbs
```diff
@@ -26,5 +25,4 @@ href="https://fonts.googleapis.com/css2?family=SUSE:wght@100..800&display=swap"
 <link rel="preconnect" href="https://www.google-analytics.com">
 <link rel="preconnect" href="https://www.googletagmanager.com">
 <script async="" src="https://www.googletagmanager.com/gtag/js?id=G-PSW07XK7TM"></script>
```
viccuad, line 29: Maybe it makes sense to parametrize this GA id here, and add it into `kw-remote-community-playbook.yml` as a variable. Then, we can overwrite it somehow when building for netlify.

viccuad (changes requested):
I think we should disable GA somehow, when deploying in CI: https://github.com/kubewarden/docs/blob/main/netlify.toml#L5

kw-community-docs-supp-files/partials/head-meta.hbs
jhkrug: Yes, let me think.

viccuad (approved):
LGTM, thanks for the effort!

PR #756, opened 2026-05-19
feat: generate static HTML redirect files for old Docusaurus URLs

viccuad:
## Description

This PR removes the Netlify redirect configuration and moves to have static files that redirect.
This allows us to keep serving the page with gh-pages.

Add scripts/generate-redirects.sh, which creates meta-refresh HTML files at old
flat paths (e.g. /reference/threat-model/index.html) pointing to the new
Antora URL structure (/admc/1.35/en/...).

Wire into both `make community-remote` and `make community-local`
targets so GitHub Pages serves the redirects without needing server-side
configuration.

Remove unneeded Netlify configuration for redirects.

## Test

- `<netlify-preview>/admc/1.35/en/reference/threat-model.html` , file exists, served directly
- `<netlify-preview>/reference/threat-model` , no file, redirected to <netlify-preview>/admc/1.35/en/reference/threat-model` , Netlify then serves `<netlify-preview>/admc/1.35/en/reference/threat-model.html`
- `<netlify-preview>/admc/ `, file/directory exists, served directly 

## Additional Information

On build, it creates the folder and file structure with html files, so one can hit the redirects:

```
$ ls -la build-local-community/site
total 2228
drwxrwxr-x 18 vic vic    4096 May 19 16:14 ./
drwxrwxr-x  3 vic vic    4096 May 19 16:14 ../
drwxrwxr-x  7 vic vic    4096 May 19 16:14 _/
-rw-rw-r--  1 vic vic    6837 May 19 16:14 404.html
drwxrwxr-x 13 vic vic    4096 May 19 16:14 admc/
drwxrwxr-x  2 vic vic    4096 May 19 16:14 disclosure/
drwxrwxr-x  2 vic vic    4096 May 19 16:14 enterprise/
drwxrwxr-x 10 vic vic    4096 May 19 16:14 explanations/
drwxrwxr-x  2 vic vic    4096 May 19 16:14 glossary/
drwxrwxr-x 30 vic vic    4096 May 19 16:14 howtos/
-rw-rw-r--  1 vic vic     553 May 19 16:14 index.html
-rw-rw-r--  1 vic vic     538 May 19 16:14 index.html.bkp
drwxrwxr-x  2 vic vic    4096 May 19 16:14 introduction/
drwxrwxr-x  4 vic vic    4096 May 19 16:14 kubewarden/
lrwxrwxrwx  1 vic vic      17 May 19 16:14 latest -> kubewarden/latest/
[...]

PR #761, opened 2026-05-22
refactor: rename doc components to improve URL SEO

flavio:
Rename source folders and Antora component names for clearer URLs:

- docs/admc/ + name: admc → docs/admission-controller/ + name: admission-controller
- docs/sboms/ + name: sboms → docs/sbom-scanner/ + name: sbom-scanner

Other changes:
- fix: build PR previews and CI tests from local checkout. Without this the netlify preview would fail (in this case, due to renames), or show stale contents (since it always pulled from `main`, not from the PR branch)
- fix: use relative path for logo home link. When clicking on the logo on the top left side of the screen, go to the root of the deployed website instead of https://docs.kubewarden.io


jhkrug:
Sure, we can do this. The playbooks will need adjusting as well. Also, maybe move docs/kw to docs/kubewarden. If you like, assign to me and I'll do all that and make sure it builds ok.

Thanks for the change to the netlify build. Yes, it should do a community-local build and use HEAD. The playbook for community-remote for gh-pages will need adjusting to the new paths. And in the community-local playbook (for building on your own workstation). Maybe I can reduce the number of playbooks back down to 2.

viccuad:
This sounds like a good change, +1 (can't formally approve, github chokes when going to the files tab for approving...)

> Maybe I can reduce the number of playbooks back down to 2.

That would also be useful.

flavio:
> Thanks for the change to the netlify build. Yes, it should do a community-local build and use HEAD. The playbook for community-remote for gh-pages will need adjusting to the new paths. And in the community-local playbook (for building on your own workstation). Maybe I can reduce the number of playbooks back down to 2.

Sorry for not mentioning that into the PR body, I just fired the PR and ran to have lunch :smile: 

I think it would be good to go down to one playbook, maybe with another PR?

flavio:
I've pushed one final commit and updated the PR body to mention all the changes. If you're fine with that I would merge this PR and then reduce the number of netlify playbooks with another PR

jhkrug:
Sure. #756 from @viccuad will also need adjusting.

PR #768, opened 2026-05-28
fix(ci): update asciidoc CLI reference documentation

jvanz:
## Description

Updates the updatecli script to copy the kwctl and policy-server CLI reference documentation generate now in asciidoc.

This PR requires the adm-controller change to be merged https://github.com/kubewarden/adm-controller/pull/1778

Fix https://github.com/kubewarden/adm-controller/issues/1751

# Test

https://github.com/jvanz/docs/pull/64/changes

PR #773, opened 2026-06-03
Script to establish a new release structure.

jhkrug:
Made use of Copilot. wdyt.

for e.g.

`make-new-release admission-controller 1.36 1.37 1.38`

It needs explicit versions provided. A future updatecli may be used to determine what versions to use based on the software tagged releases.

### DIFF
--- a/bin/make-new-release
+++ b/bin/make-new-release
@@ -1,0 +1,113 @@
+#!/bin/bash
+
+require_dir() {
+  local path="$1"
+  local label="$2"
+
+  if [[ ! -d "$path" ]]; then
+    echo "Error: '$label '$path' not found."
+    exit 1
+  fi
+}
+
+require_file() {
+  local path="$1"
+  local label="$2"
+
+  if [[ ! -f "$path" ]]; then
+    echo "Error: '$label '$path' not found."
+    exit 1
+  fi
+}
+
+run_sed_update() {
+  local path="$1"
+  shift
+
+  if ! sed -E -i.bkp "$@" "$path"; then
+    echo "Error: failed to update '$path'."
+    exit 1
+  fi
+
+  rm "$path.bkp"
+}
+
+repo_root="$(git rev-parse --show-toplevel)"
+
+if [ -z "$repo_root" ]; then
+  echo "Error: not a git docs repository?"
+  exit 1
+fi
+
+cd "$repo_root" || exit 1
+
+if [[ "$#" -ne 4 ]]; then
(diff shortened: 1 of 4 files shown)


viccuad (commented):
I think this is a great idea!

Given that it's not the first shell script that there is on the repo, I would like to suggest we put the file under `scripts/make-new-release.sh`. I would also add a CI job to run shellcheck against these files, see for example https://github.com/kubewarden/adm-controller/blob/main/.github/workflows/ci.yml#L394-L402

bin/make-new-release
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash
```
viccuad:
```suggestion
#!/usr/bin/env bash
```

So it works on more places.

jvanz:
Unfortunately, it looks like we were working in the same thing. Because we have a duplicate feature 🤭.  I've also add something similar in this [PR](https://github.com/kubewarden/docs/pull/768) using a new (and wip) updatecli [policy](https://github.com/kubewarden/updatecli-policy-antora/pull/1).

I don't have strong feeling about my changes. We can follow this PR if it's simpler. I just need to adapt my PR to use this script in the updatecli compose file.

jhkrug:
Ahh, ok. We can chat about it in the standup a bit. My motivation is a script, which when working well, can go in product-docs-common and be used across other community and product-docs docs repos. Could also be done with GH actions, I suppose.

jvanz:
> Ahh, ok. We can chat about it in the standup a bit. My motivation is a script, which when working well, can go in product-docs-common and be used across other community and product-docs docs repos. Could also be done with GH actions, I suppose.

I see. I think we can move the updatecli [pipeline](https://github.com/kubewarden/updatecli-policy-antora/blob/67fc87b08e261cdf071c275aec56355915eba070/policy/updatecli.d/antora.yaml#L1) from the policy that I shared to a policy inside the docs repository. Replacing the script there where does the same logic from your script here by a call of this script. Keeping all the rest of the steps to fetch the new version and calculating the next one. Therefore, we can remove one additional repository where we need to keep the deps updated. And you still can share the script with others.

jhkrug:
Another thing to add is updating :revdate: for a new release. Probably.

jhkrug:
Hi @jvanz sounds ok, we get everything needed from the updatecli/actions power, and the script can be developed and then moved for wider usage to product-docs-common. Wins all round.

jvanz (approved):
LGTM. @jhkrug I would say just to double check why the spell check CI is not happy, fix it and merge it. ;)

PR #776, opened 2026-06-08
Re-arrange the archived docs folder to better reflect current layout.

jhkrug:
It's just the archive, no changes to any current docs. Inside the archived_docs folder, I moved all the old kw versions into a new subdir called admission-controller. That's it.


viccuad:
I cannot approve through the UI (it says "only users with access to this repo can approve", which I have.. I think GH UI craps itself because the files tab is too big?).

LGTM.

PR #802, opened 2026-06-22
Re-arrange howtos nav a little. Group together installation howtos.

### LINKED ISSUE #795: GitOps instructions: create a dedicated section
Currently we have instructions about how to install KW admission controller using Rancher Fleet and ArgoCD. It would be great to move these pages under a "folder" in our sidebar, to make it easier to find them.

jhkrug:
May close #795


flavio:
Everything looks good to me and matches my expectation. However, reading the contents for since a long time, I wonder if we shouldn't remove [these instructions](https://deploy-preview-802--docs-kubewarden-io.netlify.app/admission-controller/1.36/en/howtos/airgap/02-install) and leave only [these ones based on hauler](https://deploy-preview-802--docs-kubewarden-io.netlify.app/admission-controller/1.36/en/howtos/airgap/03-hauler).

This is more a question for the @kubewarden/kubewarden-developers.

I personally think we should leave only the hauler ones since:
- they provide a better UX (get everything in a single sweep, verify signatures,...)
- it's the flow we constantly keep tests with our test automation

docs/admission-controller/version-1.37/modules/en/nav.adoc
```diff
@@ -73,7 +73,11 @@ include::partial$variables.adoc[]
 ** xref:explanations/architecture.adoc[{admc-short-name} architecture]
 * Howtos
 ** xref:howtos/tasks.adoc[]
-** xref:howtos/install-kwctl.adoc[Install `kwctl`]
+** Installation howtos
+*** xref:howtos/install-kwctl.adoc[Install `kwctl`]
```
jvanz, line 77: I think the kwctl installation and KubeStellar documentation should not be in this group. From what I understand from the issue, only the other documentation should be in this "folder". I think we can also rename the new group to "GitOps" or something similar.

jvanz (changes requested):

jvanz:
> Everything looks good to me and matches my expectation. However, reading the contents for since a long time, I wonder if we shouldn't remove [these instructions](https://deploy-preview-802--docs-kubewarden-io.netlify.app/admission-controller/1.36/en/howtos/airgap/02-install) and leave only [these ones based on hauler](https://deploy-preview-802--docs-kubewarden-io.netlify.app/admission-controller/1.36/en/howtos/airgap/03-hauler).
> 
> This is more a question for the @kubewarden/kubewarden-developers.
> 
> I personally think we should leave only the hauler ones since:
> 
>     * they provide a better UX (get everything in a single sweep, verify signatures,...)
> 
>     * it's the flow we constantly keep tests with our test automation

Fine by me. If we decided to do that, we need to update our CI to remove the policy and image list files from the Helm chart tarball.

flavio:
Waiting fro @viccuad to come back to decide together about what to do with regards to the old airgap installation

docs/admission-controller/version-1.37/modules/en/nav.adoc
flavio: @jhkrug :point_up: I think we can sort this out before victor comes back, since it's unrelated with the other topic

viccuad:
I'm ok with dropping the docs and imagelist.txt, policylist.txt files and related build bits.

PR #806, opened 2026-06-23
Sbomscanner docs automation

jvanz:
## Description

Add the automation to automatically bump SBOMscanner components docs version.

### DIFF
--- a/.github/workflows/updatecli.yml
+++ b/.github/workflows/updatecli.yml
@@ -11,7 +11,7 @@ permissions:
   pull-requests: write
 
 jobs:
-  updatecli:
+  admission-controller-docs-update:
     runs-on: ubuntu-latest
     if: github.ref == 'refs/heads/main'
     steps:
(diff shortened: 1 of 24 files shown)


alegrey91:
@jvanz are there any blockers for this PR?
Does this include CRD references too?

jvanz:
> @jvanz are there any blockers for this PR?

We need to agree on the doc version name directory. As far as I can see, we should rename SBOMScanner directories to: `version-MAJOR.MINOR`. Therefore, we can use the same automation used by the adm-controller. What do you think? 

>  Does this include CRD references too?

No. This is only to release a new component documentation version.

alegrey91:
> > @jvanz are there any blockers for this PR?
> 
> We need to agree on the doc version name directory. As far as I can see, we should rename SBOMScanner directories to: `version-MAJOR.MINOR`. Therefore, we can use the same automation used by the adm-controller. What do you think?
> 
> > Does this include CRD references too?
> 
> No. This is only to release a new component documentation version.

Fine to me.
@flavio @fabriziosestito wdut?

fabriziosestito:
Fine by me

jvanz:
@alegrey91 I'll open a second PR to add the automation to update CRDs later. Even if this PR is simple, the number of changed file are not so small. And, I do not want more thing in the PR. Therefore, I'll keep this one with the change only to automate the version creation.

alegrey91:
> @alegrey91 I'll open a second PR to add the automation to update CRDs later. Even if this PR is simple, the number of changed file are not so small. And, I do not want more thing in the PR. Therefore, I'll keep this one with the change only to automate the version creation.

Sounds like a good plan. Thanks ;)

docs/sbom-scanner/version-0.12/antora.yml
```diff
@@ -1,7 +1,7 @@
 name: sbom-scanner
 title: SBOM Scanner
 version: "0.12"
-prerelease:
+display: "0.12-latest"
```
jhkrug, line 4: Do we want the -latest to be displayed here? AdmissionController doesn't. Can we keep them the same either both without or both with.

jhkrug (changes requested):

docs/sbom-scanner/version-0.12/antora.yml
jvanz: The admission-controller component use the `display: "x.xx-latest` in the [latest](https://github.com/kubewarden/docs/blob/d0dd7d998b428a87ec29d81851fe1ec0d4b4f1b5/docs/admission-controller/version-1.36/antora.yml#L4) documentation version. But this is not rendered in the webpage. In the preview we can see the  same here . Is this an issue?

jhkrug: ok, i'll investigate that. But it's not a problem then. Just a bit surprising. I'll investigate outside this PR. Thanks.

jhkrug (approved):
lgtm.

PR #810, opened 2026-06-29
At @viccuad suggestion. Explict attribute to reference a security tea…

jhkrug:
…m as appropriate for the build target.

Also, a link fix and a language tweak, as I was there.

- [x] I have read and understood the [Kubewarden AI Policy](https://github.com/kubewarden/community/blob/main/AI_POLICY.md)


viccuad (commented):
I think it's a good addition, but we need to define the new `project-security-team` for it to render.

jhkrug:
Ahh, yes, in shared. I did do it then forgot to add that file as it's outside the main docs dir. Doh!  Will be ready after the meeting today.

viccuad (approved):
LGTM!

Issue #824, opened 2026-07-16
Parametrize strings for downstream consumers


viccuad:
Currently, docs.kubewarden.io have hardcoded:
- Namespace where the admission controller is installed (`kubewarden`)
- Release name where the admision controller is installed (`admission-controller`)
- etc

#### Acceptance criteria

Parametrize:
- Helm namespace
- Helm release name
- OCI repo URI (we keep the same OCI namespace)
- OCI tag (as versions may differ).

Don't parametrize the tag for auto-bumping, as that may break instructions.
Don't parametrize the OCI repo URIs of policies, until they are mirrored.

so downstream consumers can set their own.

manumathon:
I can take this up.

jvanz:
I've started to address this issue on this [PR](https://github.com/kubewarden/docs/pull/831)

flavio:
Should we close this issue now that the linked PR has been merged?

jvanz:
> Should we close this issue now that the linked PR has been merged?

We have the variables with the data. Now we need to update the documentation pages to use it. Only the quick start is using this variables

jhkrug:
Let me check through this. @jvanz do you think there are more documents that need to be changed? Now, that is apart from quick-start.adoc. There should be limited numbers of cases where attribute values (or code) needs to be different between community and product.

So, my inclination is that this issue could be closed. Thanks.

jvanz:
Ok. Fine by me. In any case, when we found a reference that could be update we can fix that on demand. Like in this [case](https://github.com/kubewarden/docs/pull/854/files)

PR #831, opened 2026-07-20
refactor(quickstart): extract quickstart text to partials

jvanz:
## Description

Updates the quick-start documentation extracting some text to partials. This allows the product documentation to override the sections that do not match community docs. Furthermore, add some variable that customize the fields that change if it's community or product version.

### DIFF
--- a/docs/admission-controller/version-1.37/modules/en/pages/quick-start.adoc
+++ b/docs/admission-controller/version-1.37/modules/en/pages/quick-start.adoc
@@ -28,7 +28,7 @@ The {admc-name} stack comprises:
   `PolicyServer`. The {admc-short-name} `PolicyServer` loads and evaluates your
   administrator's policies.
 * One or more {admission-policy} resources: policies for a defined namespace.
-* A deployment of a `{admcontroller-name}`: this controller monitors the
+* A deployment of a `{admc-release}`: this controller monitors the
   {cluster-admission-policy} resources and interacts with the {admc-short-name}
   {policy-server} components.
 
@@ -81,7 +81,7 @@ The {admc-short-name} has three main components which you interact with:
 
 === `PolicyServer`
 
-The `{admcontroller-name}` manages a {admc-short-name} `PolicyServer`. You can
+The `{admc-release}` manages a {admc-short-name} `PolicyServer`. You can
 deploy multiple {policy-server}s in the same Kubernetes cluster.
 
 A `PolicyServer` validates incoming requests by executing {admc-short-name}
@@ -123,7 +123,7 @@ Overview of the attributes of the `PolicyServer` resource:
 | N
 | `serviceAccountName`
 | The name of the `ServiceAccount` to use for the `PolicyServer` deployment. If
-  there is no provided value, the default `ServiceAccount` from the installation namespace, of `{admcontroller-name}`, gets used.
+  there is no provided value, the default `ServiceAccount` from the installation namespace, of `{admc-release}`, gets used.
 
 | N
 | `env`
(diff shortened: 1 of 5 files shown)


shared/modules/en/partials/kubewarden-variables-1.36-forward.adoc
```diff
@@ -42,6 +42,11 @@ ifeval::["{build-type}" == "community"]
 :admc-abbrev: {admc-community-abbrev}
 :intro-paragraph: {admc-name} is a xref:#_vendor_neutrality[vendor neutral], https://cncf.io[CNCF] Sandbox project, originally created by https://www.rancher.com/[SUSE Rancher].
 :project-security-team: {kw-community-project-name} team
+:admcontroller-name: admission-controller
+:default-policy-server-image: ghcr.io/kubewarden/adm-controller/policy-server
+:admc-chart-location: kubewarden/admission-controller
+:admc-namespace: kubewarden
+:admc-release: admission-controller
```
jvanz, line 53: I have the feeling that we can get ride off this `ifeval` if we change on how the partials/variables are defined. But that may require some change in the playbook and workflow already set by John. And I do not want to change that. Thus, I think we are fine to add this variables here for now. ;)

docs/admission-controller/version-1.37/modules/en/partials/variables.adoc
```diff
@@ -1,4 +1,7 @@
 // Include shared project global variables
 include::latest@shared:en:partial$kubewarden-variables-1.36-forward.adoc[]
 
-// global variables for this version below:
+:default-container-image-tag: v1.37.0
+// Fully-qualified policy server image reference. {default-policy-server-image}
+// comes from the shared file; {default-container-image-tag} is set above.
+:default-policy-server-image-tag: {default-policy-server-image}:{default-container-image-tag}
```
viccuad, line 7: Given that we will have OCI URLs for the rest of the container images and the policies, I would like to not parametrize the whole policy-server URI, but the OCI repository, the namespace, the image name, and the tag.

This would match the acceptance criteria of https://github.com/kubewarden/docs/issues/824.

docs/admission-controller/version-1.37/modules/en/pages/quick-start.adoc
```diff
@@ -28,7 +28,7 @@ The {admc-name} stack comprises:
   `PolicyServer`. The {admc-short-name} `PolicyServer` loads and evaluates your
   administrator's policies.
 * One or more {admission-policy} resources: policies for a defined namespace.
-* A deployment of a `kubewarden-controller`: this controller monitors the
+* A deployment of a `{admcontroller-name}`: this controller monitors the
```
viccuad: Shouldn't this and other changes be with `{admc-release}` instead?

viccuad (changes requested):

docs/admission-controller/version-1.37/modules/en/partials/variables.adoc
jvanz: This is just a shortcut. Just to avoid the need to remember every variable used to build the URLs. We can play with the variable. Note that this variable is composite of the other other one defined at `kubewarden-variables-1.36-forward.adoc`. We still can have individual variable for each URL part. But I like the shortcut to the container image full URL

viccuad: Fair enough, resolving.

viccuad (approved):

viccuad (approved):

jhkrug (approved):
lgtm

PR #832, opened 2026-07-21
fix(nav): conditional for product migration docs

jvanz:
## Description

Updates the nav.adoc file adding a conditional to show migration documentation specific for product documentation.


viccuad (approved):
Approving. I will rebase https://github.com/kubewarden/docs/pull/834 to make use of it.

PR #838, opened 2026-07-23
feat(variables): variables for specific docs type

jvanz:
## Description

Adds a partials file with variables that have values that are attached to the doc type and version. For now, the variable used is the container image tag that can change from community and product documentation.


viccuad (approved):

jhkrug (approved):
lgtm

PR #849, opened 2026-08-05
fix: callout to explain minimumMatches field

jvanz:
## Description

Adds a warning callout in the verification config file format to highlight for users that the minimumMatches field is related to the satisfied items in the anyOf list. Not the independent signers.

### DIFF
--- a/docs/admission-controller/version-1.38/modules/en/pages/reference/verification-config.adoc
+++ b/docs/admission-controller/version-1.38/modules/en/pages/reference/verification-config.adoc
@@ -31,6 +31,20 @@ The configuration has 2 root keys:
 * `anyOf`: You must satisfy at least `anyOf.minimumMatches` of all listed
   verification information to accept a container image as signed.
 
+[#_minimummatches_warning]
+[WARNING]
+====
+
+`anyOf.minimumMatches` counts satisfied *entries in the
+`anyOf.signatures` list*, not distinct signers. If two entries can both be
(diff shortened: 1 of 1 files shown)


flavio (approved):

docs/admission-controller/version-1.38/modules/en/pages/reference/verification-config.adoc
```diff
@@ -170,6 +183,16 @@ field is `1`.
...
 
+[WARNING]
+====
+
+Same caveat applies here: `minimumMatches` counts satisfied entries, not
+signers. If any two entries in this list can be met by the same key or
+identity, `minimumMatches: 2` doesn't guarantee two different people signed
+the resource.
+
+====
```
jhkrug, line 192: The same caveat as the earlier new warning? If so there is a large amount of text between the two so maybe a anchor and link is appropriate? I thought it just needs to be clearer as to what you are referring to. Also, is the word 'caveat' necessary, just 'the same [as linkref] applies here' will probably suffice.  

I'll set this as 'changes needed' until it's clear. Thanks.
 
```suggestion
[WARNING]
====

The same applies here: `minimumMatches` counts satisfied entries, not
signers. If any two entries in this list can be met by the same key or
identity, `minimumMatches: 2` doesn't guarantee two different people signed
the resource.

====
```

jhkrug (changes requested):
Just looking to clarify my comment. Thanks.

PR #850, opened 2026-08-07
fix(ci): do not update revdate when content has no changes

### LINKED ISSUE #818: Update `revdate` field only when content change
We have an updatecli [script](https://github.com/kubewarden/docs/blob/main/updatecli/update-cli-ref-docs.yaml) used to update the CLI references. During this updates, the `revdate` field in the header of the documentation page is updated even if the documentation content has [no changes](https://github.com/kubewarden/docs/pull/799). This needs to be fixed.

### Acceptance criteria
- Update the updatecli scripts to bump the revdate field **only** when the actual content of the documentation page change.

jvanz:
## Description

Updates a updatecli script used to update the the CLI reference documentation to avoid update the revdate field in the headers when there is no content change in the file. This field should be updated only when the documentation content actually change.

### DIFF
--- a/updatecli/updatecli.d/headers.yaml
+++ b/updatecli/updatecli.d/headers.yaml
@@ -24,6 +24,20 @@ scms:
 
 sources:
 # {{ range $index, $file := .files }}
+  # Read the ":revdate:" currently committed in the documentation repository so
+  # the header rebuilt below is byte identical when the CLI reference did not
+  # change. Otherwise every run would produce a revdate only diff and open a
+  # pointless pull request. The "revdate" target bumps the date afterwards, but
+  # only when the "commit" target really changed the file.
+  # Note: this source fails if the file has no ":revdate:" line.
+  currentRevdate{{ $index }}:
+    kind: file
+    scmid: target
+    spec:
+      file: {{ $file.path }}
+      matchpattern: '(?m)^:revdate: .*$'
+    transformers:
+      - trimprefix: ":revdate: "
   file{{ $index }}:
     kind: file
     transformers:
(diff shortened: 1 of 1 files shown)


updatecli/updatecli.d/headers.yaml
```diff
@@ -24,14 +24,28 @@ scms:
...
+      - trimprefix: ":revdate: "
   file{{ $index }}:
     kind: file
     transformers:
       - addprefix: |
           include::partial$variables.adoc[]
           = {{ $file.title }}
           :page-languages: [en, de, es, fr, ja, pt, zh]
-          :revdate: {{ now | date "2006-01-02" }}
+          :revdate: {{ source (printf "currentRevdate%d" $index) }}
```
viccuad, line 48: I wonder if we could entirely drop this line (and drop `sources.currentRevdate`), and only add the revdate in the targets, which will happen only when there's a commit because there was a change.

Then the revdate is committed, and on next run, the file transformer will leave it untouched and commit the rest of changes. Which will trigger the revdate target again to update it.

jvanz: I'm not sure if this gonna work. This pipeline is applied on top of the file [copied](https://github.com/kubewarden/docs/blob/ac63f9354f73c5a9e7bf024a0863272c1db18030/updatecli/update-cli-ref-docs.yaml#L1-L14) from the upstream repository (adm-controller) not in the current committed version. The CI always copy the latest version of the file and add the headers prefix. In other words, the file is overwritten every run. This "source" ensures that the CI will generated the same file as before if there is not other changes. 

At first glance, if we want to add the headers only when the reference docs has some changes, we need more changes in the CI. To allow updatecli compare the CLI reference docs only (with no headers). I think it's doable. But require more change that I would prefer to do it later. wdyt? :)

viccuad: aah now I understand. Thanks for giving it a thought! I'm resolving this conversation, no need to go crazy with this.

viccuad (approved):
LGTM!

PR #854, opened 2026-08-11
fix(ssc): invalid policy server spec and missing cli flags

jvanz:
## Description

Fixes some commands used in the Secure Supply Chain docs adding missing CLI flag to kubectl command. As well as, fixing a invalid PolicyServer spec.


viccuad (approved):

jhkrug (approved):
lgtm

PR #860, opened 2026-08-12
fix(cli): move CLI reference docs to partial

jvanz:
## Description

To simplify the updatecli script to copy the CLI reference documentation into the docs repository, the reference documentation content has been moved to a partial and include in the doc page. Following the same approach used to the CRDs documentation.

This remove the step of updating the revdate field for now. This can be done later and share it with the other ref docs as well. I just want first to move the content to a partial file and reuse some of the scripts already used in the CRD pipelines


viccuad (approved):
LGTM, nice simplification

jhkrug (approved):
ok

PR #883, opened 2026-09-18
glow up

flavio:
This is a restyling of the docs website, to make it look better integrated with the new main website (see https://github.com/kubewarden/kubewarden.io/pull/457).

- All the styling changes are specific to this website. It should not break downstream consumers of the docs
- I redid the landing page

Glow up time!

[Live preview](https://deploy-preview-883--docs-kubewarden-io.netlify.app/)

glow up time!


viccuad (approved):
LGTM. Lovely! Don't see anything troublesome.

jhkrug (approved):
lgtm. Very nice.

kw-community-docs-supp-files/fonts/SUSE-OFL.txt
```diff
@@ -0,0 +1,93 @@
+Copyright 2025 The SUSE Project Authors (https://github.com/SUSE/suse-font)
```
viccuad, line 1: I really like the vendoring of the fonts, this helps with offline docs in airgap.

I think we should have in mind downstream consumers when reviewing the changes, and see if there's needed changes downstream.

Had a look on what else are we missing for airgap, and it seems it's only Mermaid. For vendoring Mermaid, which I think is out of this PR, we could add it under `js/vendor/` (needs a change of the init script on downstream consumers) or to `package.json` (needs also to be done in downstream consumers).

flavio:
rebased against `main`, now runtime-enforcer docs are included! :partying_face:

kw-community-docs-supp-files/fonts/SUSE-OFL.txt
flavio: > I think we should have in mind downstream consumers when reviewing the changes, and see if there's needed changes downstream.

I did another check (not just the LLM, me too), everything should be fine since downstream consumers don't use the style folders that have been created.

> Had a look on what else are we missing for airgap, and it seems it's only Mermaid. For vendoring Mermaid, which I think is out of this PR, we could add it under js/vendor/ (needs a change of the init script on downstream consumers) or to package.json (needs also to be done in downstream consumers).

Yes, we could do that with another PR. According to the LLM there are also other small things that need vendoring (like some js from ArtifactHub)

jvanz (approved):

jhkrug:
I know I approved, but:

image

The expand contract button wibbles around when used instead of staying put.


And a minor point. Any particular reason for moving the component/version selectro to the top of the column?

flavio:
> The expand contract button wibbles around when used instead of staying put.

Fixed, I've also moved to the icon provided by our icon set.

> And a minor point. Any particular reason for moving the component/version selectro to the top of the column?

Moved back to the bottom

